PR #5, opened 2018-11-17
Allow embed with D=1

heliosdrm:
It is a trivial case, but used in code of RecurrenceAnalysis. If this should not be done, I can tweak the code in the other place.


Datseris:
I `@btime` the new and old embed, and there wasn't any problem!

PR #9, opened 2018-11-23
New methods for estimating the embedding dimension

heliosdrm:
This PR contains the following changes:

* d26e607 : improve performance of `_average_a` by not calculating `R1`, but updating the distance of the nearest neighbors of `R2` with the delayed values of the original time series.

* 3cb6fb0 : add other methods for estimating the optimal embedding dimension (Kennel's FNN -in `fnn` - and Krakovská's FFNN - in `f1nn`, I preferred using the `1` for "first" instead of yet another confusing `f`). The function `estimate_dimension` is renamed as `afnn` (standing for "average FNN"), although the original interface is kept by adding an alias with the older name.

I have modified `afnn` to accept yet another parameter - the index of the norm `p` for calculating distances, which defaults to `Inf` as in Cao's paper and the previous version of the function, but now can be changed to `2` if we want to use Euclidean distances (as in the other methods), etc.. This is combined with a new auxiliary function: `_increase_distance`, which is used to avoid calculating the embedded time series at `γ+1` (see previous commit), but works with different values of `p`.

* e5ce6c6 : just a revert of a silly change that I had made to REQUIRE.

The new version of Cao's method takes approx. 70% of the time it took before, and the other methods have approximately the same performance. None of the methods are exported in this PR, because I am not sure about what should be the interface. Options are:

*  Extend `estimate_dimension` with some optional parameter that lets you choose what method should be used. (In that case it may be advisable to revise the return types, since `fnn` returns an array of integers, whereas the other return an array of floats.)
* Keep `estimate_dimension` just as a legacy alias of `afnn`, and export the new methods.
* Some combination of the previous?

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.7.0-beta2
-Statistics
 StaticArrays 0.8
 Distances 0.7
 NearestNeighbors 0.4
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,5 @@
 julia 0.7.0-beta2
+Statistics
 StaticArrays 0.8
 Distances 0.7
 NearestNeighbors 0.4
--- a/src/estimate_dimension.jl
+++ b/src/estimate_dimension.jl
@@ -33,7 +33,7 @@ The quantity that is calculated depends on the algorithm defined by the string `
 * `"fnn"` is Kennel's "False Nearest Neighbors" method [2], which gives the
     number of points that cease to be "nearest neighbors" when the dimension
     increases. This number drops down to zero near the optimal value of `γ`.
-    This method accepts the keyword arguments `Rtol` and `Atol`, which stand
+    This method accepts the keyword arguments `rtol` and `atol`, which stand
     for the "tolerances" required by Kennel's algorithm (see [`fnn`](@ref)).
 * `"f1nn"` is Krakovská's "False First Nearest Neighbors" method [3], which
     gives the ratio of pairs of points that cease to be "nearest neighbors"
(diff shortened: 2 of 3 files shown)


heliosdrm:
Yet another thing: while analyzing the code of `estimate_dimension` (`afnn` in this PR), I found the nomenclature a bit confusing, w.r.t. the reconstructed time series `R1` and `R2`. My first impression was that `R1` would be the series reconstructed with the smaller dimension (`γ`), and `R2` would be reconstructed with `γ+1`. It took me a while noticing that it was the other way round, and I had to struggle with myself to remember it.

In `fnn` I have called the reconstructed time series just `y` (following the nomenclature of Kennel's paper). Krakovská's paper fof FFNN does not give any particular nomenclature, and in `f1nn` I have used `Rγ` for the time series reconstructed with `γ`, and `Rγ1` for `γ+1`. This was more intuitive for me. But I'm not totally comfortable with having different nomenclatures in each piece of code, even inside the same file!

Please feel free to fix the nomenclature of the different functions to whatever you think is better for readability and ease of maintenance.

Datseris:
Hey @heliosdrm , sorry I took so long. 

Thanks a lot, this looks awesome! Very nice and smart approach to increase performance.

> The new version of Cao's method takes approx. 70% of the time it took before

Nice!

Here are some comments from me:

- [ ] I agree with you about the nomenclature. I also find the `Rγ` and `Rγ1` very readable and thus I agree with this naming.
- [ ] I also fully agree that the same nomenclatures should be used through out, so `Rγ` should be used everywhere. Maybe also replace `y` to keep consistency throughout? But that is up to you.
- [ ] The two new methods also need to be tested (maybe just repeat on them the existing tests on `R_params.jl` file, modifying accordingly?)

For the design on `estimate_dimension`, here is my take:
* `estimate_dimension` name stays and is the exported name. Its call signature is changed to:
`estimate_dimension(s::AbstractVector, τ:Int, γs = 1:5, method = "afnn", p = Inf; kwargs...)` where `kwargs` is propagated to the internal functions.
* Depending on the `method` one of the three internal functions `afnn, fnn, f1nn` is called. (this is stable as I can see that all three functions return the same type).
* The docstring of `estimate_dimension` is changed to reflect these changes. The citations of the individual functions are also copy pasted to this docstring.
* My take is that the functions `fnn, afnn, f1nn` should _not_ exported (due to their very short names which are probably understood only by people that already know them). 
* However, regardless of whether we do the above or not the existing docstrings _will_ be included in the documentation page, for even more educational value, and so that users can easily see what does each method do, what keyword arguments it accepts, etc. It is also especially important because e.g. Cao's method looks for when the ratio saturates but Krakovská's method looks for when it becomes zero. These docstrings will also be hyperlinked in the main `estimate_dimension` docstring!

if we both agree with this plan, I will gladly do the changes noted by my bullet points and subsequently write the documentation page. Could you please do the checkmarks, since you are much more familiar with the source?

Datseris:
@heliosdrm is this now on my own hands ? I lost the status quo , sorry ! I see that you checked the items on the checklist but there are no commits for them. Did you forget to push ?

heliosdrm:
Hi, I started with new methods to estimate the delay, and neglected to finish this. Sorry! I'll put my hands on it tomorrow and submit the PR.

Datseris:
Absolutely no problem! I just asked because sometimes it is common to forget to push some local changes! :)

heliosdrm:
Ok, I made the changes of the checkmarks - and also some of your bullet points. Only the documentation page is still missing. Please let me know if you need something for that.

Datseris:
Awesome! I'll get to working on the documentation ASAP! :)

heliosdrm:
The last changes on `R_params.jl` introduced a mistake that made the tests fail. It is fixed on 43d09438, but the latest tag (v0.2.0) points to the failing commit.

PR #11, opened 2018-11-23
Allow `reconstruct` with γ=0

heliosdrm:
Instead of tweaking the function `embed`, modify the `reconstruct` methods (and leave `embed` as originally).

This is convenient e.g. to include in the estimation of optimal dimensions the possibility of not doing any reconstruction at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ os:
 #     - 0.7
 julia:
   - 1.0
-  - nightly
+  # - nightly
 matrix:
   allow_failures:
   - julia: nightly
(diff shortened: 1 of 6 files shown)


Datseris:
Merging when tests pass!!!

heliosdrm:
Don't know why the `if-else` in the low-level function hanged the tests. You fixed it partially in b01c39f, but that should also be modified in the multi-dimensional method

heliosdrm:
There was a mistake in the latest change to `R_params.jl`, and the tests failed. I fixed it in master, but the latest tagged version is behind that commit.

Datseris:
Thanks, I re-released to include your fix!

PR #12, opened 2018-12-17
Mutual information

heliosdrm:
The `mutualinformation` function based on histograms that is added in this PR has less requirements than the Kraskov's method that was already defined in `old_mutual_info.jl`. Using a 1,000-points sample of a Rössler system, `@btime` reports about 45 μs for each delay for which the mutual information is calculated. The time only increases linearly with the length of the series or the number of delays that are calculated. 

The histogram may be defined by the width or number of bins, or the space may be partitioned automatically using a bisection method based on the algorithm that Fraser and Swinney recommended in their seminal paper on mutual information.

It is *based* on Fraser's and Swinney's algorithm, but it's not exactly their method. They suggested to analyze increasingly finer partitions of the joint histogram between the original and delayed time series until the distribution of points is approximately uniform. On the other hand the method implemented here only uses that strategy to partition the margins of the histogram (based on the distribution of the original time series), and then the full joint histogram is analyzed. This has some advantages:

* The recursive analysis of the partitions and the tests of uniformity only have to be done once for all the delays; with F&S's method this has to be done for every delay. In the examples that I have tested, this means that F&S's method is ~ 20 times slower.

* The potential advantage of F&S's method is that it stops when the joint distribution is approximately uniform in a sub-partition. In practice, this is relevant for the empty or near-empty areas of the histogram, but the way that the joint histogram is analyzed in the function that I have implemented goes through the empty boxes without processing them, so there is no relevant penalty there.

* On the other hand, F&S's method goes into ever finer partitions when there is some substructure in the joint histogram, even if the marginal histogram is rather flat in that area. This is specially noticeable for small delays. (In the extreme case, with a null delay the original and delayed series are the same, and the joint histogram only has nonzero values in the main diagonal, no matter how small are the partitions of the space.) This makes the performance much worse for small delays - which will be normally present in the search for the optimal delay.

* The opposite can also happen: I have stumbled upon cases (with greater delays) where the uniformity test is positive at the first partition of the joint histogram using F&S's algorithm, and the mutual information accidentally drops to zero. This type-II error may be expected to happen more frequently as the number of analyzed delays increases, since the the tests must be repeated for every delay.

* Using a joint histogram of fixed partitions ensures that the marginal entropy that underlies the calculation of mutual information does not change. So the values are more stable (the mutual information must be a value between zero and the marginal entropy).

In any case, I have also left the code for the original implementation of Fraser and Swinney's algorithm in an extension of `old_mutual_info.jl`, plus some code for the calculation of mutual information with Kernel Density Estimators (Moon et al., 1995, Phys Rev. E 52(3):2318-21).

Some things to do:

* Decide what interface we want to use for these functions. `estimate_delay.jl` had a function that always started at `τ=1` and looked for the first zero or local minimum until a maximum delay value. But I have made an interface for `mutualinformation` similar to that of the function to calculate the optimal dimension. - I also prefer it, since you don't need to calculate the result for all the possible delays between 1 and `maxtau`, and you may use another search criterion instead of the first local minimum.
* Write tests (which will depend on the interface, of course).

Besides, since the exact algorithm that is implemented is not published, perhaps it might be worth to spend some time in writing a paper on it? This field is not my speciality, so I leave this open question to your consideration.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,4 +3,4 @@ StaticArrays 0.8
 Distances 0.7
 NearestNeighbors 0.4
 StatsBase 0.24
-KernelDensity 0.5
+
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,3 +3,4 @@ StaticArrays 0.8
 Distances 0.7
 NearestNeighbors 0.4
 StatsBase 0.24
+KernelDensity 0.5
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -120,7 +120,8 @@ end
 Dataset{D, T}() where {D,T} = Dataset(SVector{D,T}[])
 
 # Identity constructor:
-Dataset(x::Dataset) = x
+Dataset{D, T}(s::Dataset{D, T}) where {D,T} = s
+Dataset(s::Dataset) = s
 
 ###########################################################################
 # Dataset(Vectors of stuff)
--- a/src/estimate_delay.jl
+++ b/src/estimate_delay.jl
@@ -1,4 +1,5 @@
 using StatsBase: autocor
+using KernelDensity
 # using LsqFit: curve_fit
 export estimate_delay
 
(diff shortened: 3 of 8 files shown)


heliosdrm:
Also: talk with the owner of [InformationMeasures.jl](https://github.com/Tchanders/InformationMeasures.jl). That only uses fixed-bin histograms, but it has other interesting variations, and it would be nice if we can collaborate.

Datseris:
I've just pushed a commit that provides the new interface. How does it look?

heliosdrm:
"mi_zero" can be removed: the mutual information should always be positive, and only exceptionally may become zero.

An alternative would be exporting `autocor` and `mutualinformation`, and make an independent function that looks for the first zero or the first minimum. A nice way to use them together is with a pipe, e.g. if the function is `firstmin(x::Vector{T}) where {T}`:

```
mutualinformation(s, τs) |> firstmin
```

Ah, and one detail: this interface assumes that `τs` is sorted. This is pretty obvious, but should be documented.

Datseris:
Although having a `firstmin` function sounds more general, at the moment I can't think of any other immediate use. We can implement what you suggest, but I'd vote for `estimate_delay` to stay... Sure, I'll remove `mi_zero` and document the sorting on `τs`. 

I only use the pipe on interactive use and only following very fast functions (although I remember that this new mutualinformation is super fast when compared to the old one we had).

PR #14, opened 2019-01-04
add metric support for `estimate_dimension`

### LINKED ISSUE #10: Metric used for the estimation of dimensions
There is an inconsistency in the metrics used for calculating the optimal dimensions of `reconstruct` / `embed`.

In the code of the internal function `_average_a` (used by `estimate_dimension`), the distance `δ` of nearest neighbors is calculated using an infinity norm, as in Cao's paper - e.g. `δ = norm(R2[i]-R2[j], Inf)`. However, the `KDTree` that was used to find the nearest neighbors is calculated with the default options (`tree2 = KDTree(R2)`), including Euclidean distances (as said in https://github.com/KristofferC/NearestNeighbors.jl).

The `KDTree` should be calculated with the `Chebyshev` metric to be consistent. Another option (which I advocate) is let the user choose what `Metric` is used. PR #9 includes code to facilitate this (only for the calculation of `δ`, not for the `KDTree`).

The authors of the other methods (FNN, FFNN) used Euclidean distances in their papers. This is the only option consistent with Kennel's equations [4-5] (FNN), but in the case of the FFNN, other metrics could be used without problem.

Datseris:
Closes #10

I had to re-work a bit the code to not use `p` at all and dispatch on the metric, but I hope I did everything correctly! 

Helios let me know if you feel confident that `DelayEmbeddings` can move to 1.0 after this PR!


heliosdrm:
Looking at the code (I have made no tests myself) it looks ok, except for a small detail:

The documentation says that the metric defaults to `Euclidean()`, but in `afnn` it defaults to `Chebyshev()`. That's coherent with the metrics used in the papers (Cao uses the maximum norm), but it should be told explicitly. Or if we prefer more homogeneous defaults, use `Euclidean()` everywhere but include a note on the departure from Cao's algorithm.

Datseris:
`estimate_dimension` does default to `Euclidean()` for all methods. `afnn` defaults to `Chebyshev` but only when called directly, which is not part of the high level API. So the "standard" call to Cao's algorithm will use `Euclidean`. I made it this way because I wasn't feeling sure whether we should depart from the way Cao's paper has the algorithm...

We can add method-dependent defaults, so that `afnn` defaults to using Chebyshev, or we can leave it as is and add a note in the docstring. Which one do you prefer?

heliosdrm:
Cao's method is not strongly bound to the maximum norm. The paper says (next to eq. (1)):

*" ||·|| is some measurement of Euclidian distance and is given in this paper by the maximum norm"*

So his using the maximum norm seems to be just a choice for the examples of the paper. There is nothing else in the method that depends on using a specific metric.

For the sake of consistency I think that the default should be the same regardless on how the method is called. And I also prefer to use the default metric to be the same for all methods. Since `fnn` only makes sense with the Euclidean metric, and the publication that presents `f1nn` also uses the Euclidean metric, I think that the better choice is `Euclidean()` for all, with a note in the docstring of `afnn`.

PR #22, opened 2019-06-14
Add a plot recipe for Plots.jl and Datasets

asinghvi17:
Quality of life improvement mainly :)


src/dataset.jl
```diff
@@ -206,6 +208,14 @@ end
 Base.show(io::IO, ::MIME"text/plain", d::AbstractDataset) = print(io, matstring(d))
 Base.show(io::IO, d::AbstractDataset) = print(io, summary(d))
 
+#####################################################################################
+#                                     Plotting                                      #
+#####################################################################################
+
+# Support the plotting of Datasets-as-Matrices
+RecipesBase.@recipe f(::Type{T}, vals::T) where {T<:AbstractDataset} = Matrix(vals)
```
Datseris: Should here `AbstractDataset` have constrain to dimension of only 2?

How does `Plots` plot stuff of dimension more than 2? (i.e. what does a matrix with 5 columns mean for Plots?)

asinghvi17: All matrices decompose into a timeseries by default (so, if you do `plot(somematrix)`).  However, if you do something like `surface(xs, ys, matrix)`, then the matrix is assumed to be a grid of heights (z-values).

asinghvi17: So there's really no reason to constrain it.  If you wanted the actual time values (from a trajectory for example), you'd do `plot(LinRange(Tmin, Tmax, size(dataset, 1)), dataset)`.

asinghvi17: That was why I was advocating for the `Dataset` to function like the `DiffEqArray`, BTW - with metadata, there's a lot of cool stuff that can be done with it.

Datseris:
I accept this; can you please add a note in the changelog and increment the version in `Project.toml`?

asinghvi17:
Done

PR #28, opened 2020-03-21
Generalized delay embedding

Datseris:
This adds the genralized embedding struct `GeneralizedEmbedding` as well as the function `genembed`. The docstring of `genembed` describes the functionality, that allows us to make an embedding of a given dataset **by choosing arbitrary delay times (positive/negative) as well as arbitrary timeseries from the dataset, allowing duplicates**.

@kahaaga is this embedding adequate in its entirety for _any_ situation regarding transfer entropy calculation?

I would really, really appreciate it if you could do a quick test with your implementations on whether using this embedding solve all problems. I created this embedding for a completely independent project, (unified embedding approach by Pecora et al), but I realize it solves the problem of having this extra Julia package "generalized embeddings".


Datseris:
@kahaaga I thought about this more, and I still believe it solve all of our problems and robustly supports all Transfer Entropy related embeddings, as you can give it a 2-dim dataset and you can still get future of x, past of x and past of y. 

So I'll go ahead and merge this.

kahaaga:
@Datseris 

Hi! Sorry for the incredibly late reply. Yes, this looks perfect! I'll scrap my unfinished implementation and use this for the CausalityTools algorithms instead. Thank you so much!

PR #29, opened 2020-03-22
[WIP] Unified approach to delay coordinates, Pecora et al.

### LINKED ISSUE #17: Feature request: Pecora method
There is a great idea for time delay embedding that outperforms the simple approach. It was described in Pecora et al: _A unified approach to attractor reconstruction_, Chaos **17**, 2007 ([http://dx.doi.org/10.1063/1.2430294](http://dx.doi.org/10.1063/1.2430294)). 

It would be nice to have this algorithm included.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/68977440-feature-request-pecora-method?utm_campaign=plugin&utm_content=tracker%2F115474986&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F115474986&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Datseris:
This is work in progress and requires #28 to be merged (it already uses #28 here, so the code is runnable).

Closes #17 

I am trying to implement the paper 

> Pecora, L. M., Moniz, L., Nichols, J., & Carroll, T. L. (2007). A unified approach to attractor reconstruction. Chaos: An Interdisciplinary Journal of Nonlinear Science, 17(1), 013110. https://doi.org/10.1063/1.2430294

Unforunately, I found their method vastly inferior with anything else that exists, as the authors do not even attempt to provide the user with an answer to the question: "Well, I got the ⟨ε★⟩ , now which τ and j should I choose???", as ⟨ε★⟩ increases with increasing τ. 

Some TODOs:

- [x] Average resulting ⟨ε★⟩ over several `k` neighbors (trivial to do)
- [x] Theiler window
- [ ] Implement the Γ undersampling statistic


Datseris:
@hkraemer running the test file `/test/test_pecora.jl` gives me the following two figures:

pecora_1

pecora_2

The first one tests the data you gave me as text file. It is quite similar with your results, but not identical. Maybe because of the different way I compute `⟨ε★⟩` , and/or the different `k` we used?

The second figure is application to multiple timeseries. The cool thing is, the same source code is used for both, and it takes less than a second to run.

Datseris:
I just averaged the result over all k from 5 to 13 (all k given in the table) and the results are:

pecora_3

pecora_4

I don't see any _qualitative_ difference.

Datseris:
It is **astonishing** that the source code for this is 58 lines of code!!! Of course without continuity statistic.

hkraemer:
Yeah, it looks similar, which is good! How long was the time series you used? Did you use all the 64,000 values?  The figures will change anyway depended on the chosen tau’s . So how did you pick the tau’s from the different epsilon-star-curves? 

As I mentioned: I would take the maximum of all the epsilon’s for the different k’s and don’t average them. 

>  The cool thing is, the same source code is used for both, and it takes less than a second to run.
> 
> 
WHAT??? Julia is just… I still don’t get it. 

Did you also include the exclusion of temporal neighbours?
-----------

K. Hauke Krämer 
PhD Candidate
Email: hkraemer@pik-potsdam.de, hkraemer@uni-potsdam.de 

Research Domain 4 - Complexity Science
Machine Learning, Nonlinear Methods and Decision Strategies
Potsdam Institute for Climate Impact Research 
Telegraphenberg A 56, Room 3.13 
P.O. Box 60 12 03, D-14412 Potsdam 

Institute of Geosciences
Haus 29, Room 1.47
Karl-Liebknecht-Str. 24-25, D-14476 Potsdam-Golm

hkraemer:
I also realized that your continuity staistic’s total values are less than the one’s of the ones shown in the paper and also from my implementation. Did you normalize the input time series?
-----------

K. Hauke Krämer 
PhD Candidate
Email: hkraemer@pik-potsdam.de, hkraemer@uni-potsdam.de 

Research Domain 4 - Complexity Science
Machine Learning, Nonlinear Methods and Decision Strategies
Potsdam Institute for Climate Impact Research 
Telegraphenberg A 56, Room 3.13 
P.O. Box 60 12 03, D-14412 Potsdam 

Institute of Geosciences
Haus 29, Room 1.47
Karl-Liebknecht-Str. 24-25, D-14476 Potsdam-Golm

Datseris:
Hi Hauke @hkraemer 

1. I also noticed the smaller ε values. I did not nornalize input timeseries, did you?
2. I have not implemented exclusion of temporal neighbors yet. It is in my todos
3. You say to not average ε for different input k (the amount of neighbors inside the δ-ball), but to use the largest one. But then it makes no sense to do the same computation for various different `k`, but only for the largest `k`. ε is a strictly increasing function of `k`, and thus I only need to do the computation for the largest `k` if I want the largest ε.

hkraemer:
Hey George,

- I normalized the time series to zero mean and unit variance.

- " ε is a non-decreasing function of k," - not necessarily. Considering an additional neighbour could mean that its distances in the 1-dimensional epsilon-room, i.e. the difference of its last embedding component with respect to the last component of the fiducial point, is smaller than for the previously considered neighbours. By allowing another point to fall outside the minimum epsilon range, this would mean that the computed epsilon-min is in fact smaller for a larger k. Because there is no reason to prefer any choice of k, I suggest to take the maximum of all epsilon(k).

Datseris:
Yes you are right, I got confused. I'll change it tomorrow.

Datseris:
Hi Hauke, I finally finished the god damn method. I am now Furious that the authors never show the undersampling statistic Γ for something reproducible like the lorenz system. Only for this "lobster", which I don't see anywhere in the paper.

Anyway, the full method for all 3 lorenz timeseries as input, with all the jazz, takes around 90 seconds.  This is 3 orders magnitude slower than _just_ computing the continuity.

Here are some results for the Lorenz system (the outpuit of the file `tests/test_pecora.jl` : 

Figure_1

Figure_2

You also had some figures like these right? Can you share the undersampling statistic? I am really not sure whether my code is correct. It is an extremely complicated algorithm, and I have nothing to compare it with to know if it is good or not.

Datseris:
Here is a case where input embedding is 3 dimensional 

Figure_3

PR #36, opened 2020-06-16
Uzal cost function

Datseris:
Cost function of Uzal et al. Contributed by Hauke Kraemer.


src/uzal_cost.jl
```diff
@@ -0,0 +1,116 @@
...
+  point itself, and no temporal neighbors.
+* `Tw = 40`: The time horizon (in sampling units) up to which E_k^2 gets computed
+  and averaged over (read algorithm description).
+
+## Description
+tbd.
+
+[^Uzal2011]: Uzal, L. C., Grinblat, G. L., Verdes, P. F. (2011). [Optimal reconstruction of dynamical systems: A noise amplification approach. Physical Review E 84, 016223](https://doi.org/10.1103/PhysRevE.84.016223).
+"""
+function uzal_cost(Y; Tw::Int = 40, K::Int = 3, w::Int = 1, SampleSize::Float64 = .5,
```
Datseris, line 37: CamelCase naming in Julia is used for either Types or Modules, but not for variables. `SampleSize` should be made `samplesize`.

test/uzal_cost_test.jl
```diff
@@ -0,0 +1,201 @@
+using DynamicalSystemsBase
```
Datseris, line 1: This test file needs to be reworked into proper unit tests.

1. It cannot be using plotting packages (but the plotting code can remain, but it has to be commented out)
2. There is no reason to save and load text files, instead `DynamicalSystemsBase` can be used to produce the timeseries on demand.
3. Unit tests are performed as follows: use the `@test` macro and try to test the the returned `L` number matches something expected, either by theory or by existing literature. If you have a look at the rest of the test files it will help you structure the tests accordingly.
4. The test file can probably be shortened, at the moment it is about 200 lines but I think we can do it in 50 or so.

test/uzal_cost_test.jl
```diff
@@ -0,0 +1,201 @@
...
+using DelayEmbeddings
+using StatsBase
+using DelimitedFiles
+using Revise
```
Datseris, line 5: Revise can also be removed from the tests

test/uzal_cost_test.jl
Datseris: Ah, also the text file `roess_ts.txt` has to be deleted.

Datseris:
After testing this I've noticed that the function does not work properly with vector data. Passing a standard Julia `Vector` of 10_000 elements gave me:
```
┌ Warning: You are attempting to make a Dataset of a matrix with more columns than rows.
└ @ DelayEmbeddings C:\Users\datse\.julia\dev\DelayEmbeddings\src\dataset.jl:186
```
which is an indication of using the vector as row vector instead of a column vector (its intended use). I also noticed that when loading the saved Roessler text data, they were also in row form instead of column form, and maybe that is what made this initial design this way?

Datseris:
Running the uzal code for the Roessler system gives:

image


the Fig 7b of the Uzal paper is:

image


The two figures are qualitative the same and thus I don't worry much. I've notice that the scale is slightly different: in the Uzal paper e.g. for k=3,4 the main peak at t_w = 50 exceeds the value 1, while in our plot it stays well below 1. Is this something to care about?

src/uzal_cost.jl
hkraemer: Fair enough, will do.

test/uzal_cost_test.jl
hkraemer: Thanks for the comments. `roess_ts.txt` was added just for comparison reasons to my Matlab implementation. Regarding the expectation value of the L-statistic: I don't see a way of an independent test, i.e. an analytical reference result. The only thing I know are rough values from the Paper itself and standardized outputs from this implementation (and from my Matlab implementation).

Datseris: That is fine, if this is the best we have this is what we do! You can use the following syntax: `@test Lmin < L < Lmax` for some pre-defined, empirical minimum and maximum values.

Perhaps you also want to run it with a vector of ones, `v = ones(10000)` and see what it does.

hkraemer: Yah, I will do. This is also related to your comment below: The total value of `L` depends on the time series itself (i.e. a normalization will yield different values, but does not change the qualitative curve, when plotted against `t_w`, for instance. It also slightly depends on the prediciton horizon `Tw`. So they did not comment on this in Fig.7 in the paper. That is why I am not really concerned about the different scale. Another thing is the assertion of a column vector Dataset input. I could of course assert the input as an `AbstractDataset`-type, but how can I assert, that this is a column-vector then (execpt from manually checking the `size` and possibly transpose)?

hkraemer: Actually, I need to refine my last comment. The scale itself is very important, when comparing the minima of course. The rank of the minima, with respect to the total (negative) value, must be preserved, i.e. the same as shown in the Paper. So I think/hope, it is all good :)

Datseris: Okay this is convincing for me. 

> Another thing is the assertion of a column vector Dataset input. I could of course assert the input as an AbstractDataset-type, but how can I assert, that this is a column-vector then (execpt from manually checking the size and possibly transpose)?

I did not understand this part... can you please try to explain it differently? WHat I can say for sure is that manually checking anything and transposing anything should definitely be avoided.

hkraemer: The function works properly, when the input phase space trajectory `Y` is a Dataset "filled" with column-vectors. Meaning, the number of columns represent the dimensionality of the phase space trajectory and the number of rows the number of observations. I somehow need to assert, that `Y`is really that kind of Dataset. So I could assert this, as far as I understand within the definition of the function, typing `Y::AbstractDataset`. But then it is still not guaranteed, that it is filled with column-vectors. - So one would need to check this within the function body. Moreover, a normal Julia-`Array`, would also be fine, then one would need to convert it into a `Dataset` within the function body. Maybe there is an elegant way to assert this?

Datseris: The `AbstractDataset` type guarantees that you have column number = dimensionality. The assertion is enough.

Datseris: `Dataset` is not a `Matrix`, internally it is a very different thing.

hkraemer: Perfect!

Datseris: but then you have to change the documentation string, because it says that it can get as an input either timeseries or Dataset, while in reality it can only get datasets.

src/uzal_cost.jl
```diff
@@ -8,14 +8,15 @@ export uzal_cost
 """
     uzal_cost(Y; kwargs...) → L
```
Datseris, line 9: I think it is best to denote the expected input type in the call signature:
```
    uzal_cost(Y::Dataset; kwargs...) 
```

src/uzal_cost.jl
```diff
@@ -8,14 +8,15 @@ export uzal_cost
...
 Compute the L-statistic `L` according to Uzal et al.[^Uzal2011],
-for a phase space trajectory `Y` (timeseries or `Dataset`). The L-statistic is
+for a phase space trajectory `Y` (`Dataset`). To convert an `Array` into Dataset
+format, simply use `Dataset(Array)`. The L-statistic is
```
Datseris, line 12: THere is no reason to explain here how to convert an `Array` in to a dataset. the documentation of `Dataset` takes care of that.

src/uzal_cost.jl
```diff
@@ -30,16 +31,38 @@ an irrelevance measure.
   and averaged over (read algorithm description).
 
 ## Description
-tbd.
+The `L`-statistic based on theoretical arguments on noise amplification, the
```
Datseris, line 34: Thanks a lot for taking the time to write a description for the `L` statistic. It is very important for the spirit of DynamicalSystems.jl :) 

Do you think you can add a sentence about "what" it actually is, in the opening paragraph? (.e.g "it is a measure of the prediction/predictability", or something like that, I actually do not know...)

test/uzal_cost_test.jl
```diff
@@ -1,91 +1,70 @@
...
+using Statistics
 
-pygui(true)
+println("\nTesting uzal_cost.jl...")
 
-## Simple Check on Lorenz System
-lo = Systems.lorenz()
-
-tr = trajectory(lo, 100; dt = 0.1, Ttr = 10)
+## Check on random vector
```
Datseris, line 9: to be on the safe side, its probably worth it to add a 
`Random.seed!(1516578735)` here.

Datseris:
@hkraemer great, I am ready to merge this!!! There is only small thing left. I had to change the number of trials in the tests, because they were taking quite some time (and there is only a small limit of time we can test on the online CI for free). 

I've adjusted all tests except the Lorenz last tests. I am not sure how you chose these values there. Could you please adjust the values to some values so that the final tests pass? And then we are good to go!

hkraemer:
> @hkraemer great, I am ready to merge this!!! There is only small thing left. I had to change the number of trials in the tests, because they were taking quite some time (and there is only a small limit of time we can test on the online CI for free).
> 
> I've adjusted all tests except the Lorenz last tests. I am not sure how you chose these values there. Could you please adjust the values to some values so that the final tests pass? And then we are good to go!

Hey @Datseris, I adjusted the tests now, but had to increase the number of trials slightly, since the statistics need to cover the global minimum I want to test against. Hopefully that is still in the time limit? If not, I have to arrange it differently somehow. Let me know.

Datseris:
@hkraemer it's okay, it runs sufficiently fast now. I've added a CHANGELOG entry and incremented the project version in the `Project.toml` file, so that we can release this new function immediatelly after merging the PR.

However, when I run the tests locally, some tests still fail even after you changes...

Datseris:
I get:
```

Test Failed at C:\Users\datse\.julia\dev\DelayEmbeddings\test\uzal_cost_test.jl:195
  Expression: tau_min < min1 < tau_max
   Evaluated: 17 < 10.0 < 21
ERROR: LoadError: There was an error during testing
in expression starting at C:\Users\datse\.julia\dev\DelayEmbeddings\test\uzal_cost_test.jl:195
```

hkraemer:
This is strange, it works perfectly on my machine, and I tried many times.
Bildschirmfoto 2020-07-09 um 11 38 55

Datseris:
I'll have a closer look

PR #38, opened 2020-07-06
Garcia almeida

hkraemer:
Implementing the Garcia & Almeida Method for non-uniform (multivariate) embedding, based on the two papers: 
- Garcia & Almeida, PRE 71 (2005) https://doi.org/10.1103/PhysRevE.71.037204
- Garcia & Almeida, PRE 72 (2005) https://doi.org/10.1103/PhysRevE.72.027205


hkraemer:
The results for the N-Statistics do not match the results shown in the paper. I double-checked that. @pucicu and Yoshito Hirata implemented the method independently, gaining the same results. Hirata pusblished results based on this method in

- Hirata et al., PRE 74 (2006) https://doi.org/10.1103/PhysRevE.74.026202

I had a few discussion with @pucicu about this and finally wrote Prof. Almeida an Email for clarification. - No answer yet.

Datseris:
@hkraemer sounds good. When you feel like this is ready for review go ahead and add me as a reviewer. You can click the button on the top right of this screen, next to the first post.

Datseris:
@hkraemer before reviewing this can try to make the unit tests proper like in the Uzal PR, and remove unnecessary files etc.? Plotting code can be left in but commented out.

hkraemer:
I believe I did :)

Datseris:
@hkraemer is this ready for review? I see that the first function is unfinished:
```julia
function garcia_almeida_embed()
    ####
    ####
end
```

hkraemer:
Same with the Nichkawde method. I just finished the core algorithm, not the automatic embedding process (that is what we also did with the Pecora-method).)
-----------

K. Hauke Krämer 
PhD Candidate
Email: hkraemer@pik-potsdam.de, hkraemer@uni-potsdam.de 

Research Domain 4 - Complexity Science
Machine Learning, Nonlinear Methods and Decision Strategies
Potsdam Institute for Climate Impact Research 
Telegraphenberg A 56, Room 3.13 
P.O. Box 60 12 03, D-14412 Potsdam 

Institute of Geosciences
Haus 29, Room 1.47
Karl-Liebknecht-Str. 24-25, D-14476 Potsdam-Golm

src/garcia_almeida.jl
```diff
@@ -0,0 +1,222 @@
+using Neighborhood
+using NearestNeighbors
+using StatsBase
+using Distances
+
+export garcia_almeida_embed
```
Datseris, line 6: `garcia_albeida_embed` is not yet ready for export.

Manifest.toml
```diff
@@ -0,0 +1,118 @@
+# This file is machine-generated - editing it directly is not advised
```
Datseris, line 1: The manifest file must not be included in a packge.

hkraemer:
Hey @Datseris,
I tried to incorporate the main ideas you gave in the nice review of #45 . Due to the same structure of the main function `garcia_almeida_embed` and `MDOP` (in #45 ) there also arise similar problems. While I could make the core function `garcia_embedding_cycle` type-stable, `garcia_almeida_embed` is not. Maybe it is a problem that one does not know the final embedding dimension, i.e. the size of the returned `Array`'s and `Dataset`'s _a priori_?

Datseris:
For the moment let's focus on merging #45 first, and then work on this one, because we will also have to rebase this PR to the master branch anyways! Don't worry #45 is almost there, we only miss the profiling and performance benchmarking step!

Project.toml
```diff
@@ -8,6 +8,7 @@ Distances = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 NearestNeighbors = "b8a86587-4115-5ab1-83bc-aa920d37bbce"
 Neighborhood = "645ca80c-8b79-4109-87ea-e1f58159d116"
+Peaks = "18e31ff7-3703-566c-8e60-38913d67486b"
```
Datseris, line 11: Why is it necessary to use `Peaks` here?

hkraemer: No, we need to look for the first local minimum (in `garcia_embedding_cycle!`).

Datseris: Yes, but the method you use just compares `x[i-1] > x[i] && x[i] < x[i+1]` correct? No reason to use an external package for such a simple function. Are you using a more complicated method that does e.g. gaussian smoothening first or something else?

hkraemer: No, but there might be ideas to restrict the minima-search further in the future, i.e. min. peak distances or something like that. Also a smoothing first would be desirable I guess in order to avoid picking every mini-up-down. So far the n-statistics I had a look on were kind of smooth, so there hasn't been the need for it right now. If you think it would be way to much overhead then we do it your way. But this would be trickier to extend, if one wishes. I don't care so much to be honest, because this method does not give the promised results anyway.

Datseris: I still don't understand from your answer what does `Peaks` do. Does it do the simple check of finding the minimum `i` that satisfies `x[i-1] > x[i] && x[i] < x[i+1]` or not? Can you justify why we need it?

hkraemer: > [...] so there hasn't been the need for it right now. If you think it would be way to much overhead then we do it your way. 

> I don't care so much to be honest, because this method does not give the promised results anyway.

Again, if one would like to improove this algorithm further it would be handy having used `Peaks` in the first place. Right now, as far as I can see, it takes the 1st minimum which satisfies `x[i-1] > x[i] && x[i] < x[i+1]`

Datseris: > Again, if one would like to improove this algorithm further it would be handy having used Peaks in the first place. 

I don't agree with this statement, because I don't believe that `Peaks` will be improved with more functionality in the near future. 

> Right now, as far as I can see, it takes the 1st minimum which satisfies `x[i-1] > x[i] && x[i] < x[i+1]`

Then we should drop the `Peaks` dependency and write the 3-lines of code function that does this operation. Pre-1.0 dependencies are a very heavy price to pay and should be avoided when it is easy to do so...

hkraemer: > Then we should drop the Peaks dependency and write the 3-lines of code function that does this operation. Pre-1.0 dependencies are a very heavy price to pay and should be avoided when it is easy to do so...

alright.

src/garcia_almeida.jl
```diff
@@ -242,3 +241,37 @@ function garcia_almeida_break_criterion(FNNs, counter, fnn_thres, max_num_of_cyc
...
+            end
+        end
+        if s[i-1] > s[i] && s[i+1] == s[i]
+            flag = true
+            first_point = i
+        end
+    end
+    # make sure there is no empty vector returned
+    if isempty(minimas)
+        _, minimas = findmin(s)
```
Datseris, line 274: `minimas` is initialized as a vector of integers at the start of the function. Here `minimas` becomes an `Int` instead of vector, which makes the return type of the function type unstable. Is the return type `Int` or `Vector{Int}`? The simple change of `_, minx = findmin(s); return [minx]` ensures that always a vector of integers is returned. 

I have a better question however. Since you only need the first minimum, why do you bother collecting all minima instead of just stopping after the first and returning the index of the first minimum? You anyways do ` τs[min_idx[1]]` in the above function...

Datseris (changes requested):

src/garcia_almeida.jl
hkraemer: Hi @Datseris,

> minimas is initialized as a vector of integers at the start of the function. Here minimas becomes an Int instead of vector, which makes the return type of the function type unstable. Is the return type Int or Vector{Int}? The simple change of _, minx = findmin(s); return [minx] ensures that always a vector of integers is returned

Thank you for this valuable comment. I changed it. 

> I have a better question however. Since you only need the first minimum, why do you bother collecting all minima instead of just stopping after the first and returning the index of the first minimum? You anyways do τs[min_idx[1]] in the above function...

You are right. Since I'll need to find all maximas in a time series for my other idea regarding the Pecora-method, I set up this code, to be easily reusable for this task as well.

Datseris: > You are right. Since I'll need to find all maximas in a time series for my other idea regarding the Pecora-method, I set up this code, to be easily reusable for this task as well.

Okay, that is good idea then

PR #40, opened 2020-07-09
Nichkawde method

hkraemer:
Here the method for optimal state-space reconstruction using non-uniform time-delay-embedding proposed by Nichkawde has been implemented.

- Nichkawde, Chetan (2013), Optimal state-space reconstruction using derivatives on projected manifold, PRE 87, 022905


hkraemer:
The results shown in the paper for the Mackey-Glass-System can be approximately reproduced. I think the little change comes from different time series. For instance, my Matlab-Implementation (which is doing the same as this implementation, but about 100 time slower) produces this one here:

mackey_glass

, while this implentation produces this (just 1st embedding cycle shown):

mackeyglass_julia

I think it's alright. More tests (with the other examples in the paper) follow.

Datseris:
I've just went through this now! I have two comments:

1. Can we use another test system (maybe from the paper) that is ODE instead of DDE? For the only reason that installing and compiling DelayDiffEq is a big hit to the test run time.
2. Seems to me that we can finish the function `MDOP` by giving the total cycle number, and total `τ` range to search, and then from each curve obtain the global maximum: first obtain all local maxima, and then pick the highest of them. What do you think? So that the "high level" aspect of this function is finished already.

Datseris:
Ah, another, important comment: Only one version `::Dataset, ::Vector` should exist. Univariate timeseries are Vectors in the DynamicalSystems.jl ecosystem, offering different alternatives expands the API complexity without gain.

Datseris:
@hkraemer sorry that my review is separated into several comments. Here come more points:

* `τ_max` keyword should be changed to `τs = 1:50`. This allows the user to explicitly provide possible τ values to compute β for.
*  I don't understand why the returned `β` is a 1x100 matrix, instead of a standard 100 element vector.

hkraemer:
Can you think of any situation where you would not compute the β-statistic for a range of delay values starting from `0:τ_max`? Why would you do that? I agree on all other comments.

Datseris:
If the timeseries comes from a continuous dynamical system that is densely sampled, you can typically get away with `1:2:τmax` or even `1:4:τmax` :)

hkraemer:
okay, I am convinced. I'll change it.

hkraemer:
When I run the function,  the returned β is a normal Array{Float64,2} - type. Is this not a "standard" vector in Julia?

Datseris:
> Is this not a "standard" vector in Julia?

Nope, a standard vector is just `Vector{Float64}`, which is created via e.g. list comprehension `[x^2 for x in 1:1000]` or by broacasting over existing vectors `x = x.^2 .+ y`, or by initialized as zeros and then adding values `x = zeros(10); x[1] = 5`.

hkraemer:
If I preallocate a variabel `x = zeros(10)`, this is a 10-element Array{Float64,1} and that does not change when I fill this array with numbers/Integers etc.. I am running on Julia 1.3.1.
What am I missing?

hkraemer:
Okay, it is all done. The problem with the WINDMI-system for testing the algorithm is that they do not provide any data in the paper we could compare the results to. They just show one reconstruction picture for it and in Table 1 they provide results for the whole MDOP-procedure, i.e. embedding dimension and chosen delays. For testing the whole method I will then compare to Table 1 - data and use the WINDMI system, but for making sure the core algorithm (`β`) works as expected, I think we have to stick to the examples they presented.

src/MDOP.jl
```diff
@@ -0,0 +1,103 @@
...
+function MDOP(Y...)
+    #tbd
+
+end
+
+
+"""
+    beta_statistic(Y::Dataset, s::Dataset(::Array); kwargs...) → β
+Compute the β-statistic `β` for input phase space trajectory `Y` (of type
+`Dataset`) and a univariate time series 's' (1-dim. `Array` or `Dataset`)
```
Datseris, line 25: Here some changes I did were reverted. E.g. `s::Vector`.

hkraemer: Hmpf, sorry. I deleted the wrong function with your corrected docstring. Now it should be okay again.

Datseris:
Hi @hkraemer have a look at my latest commit. There are some helpful points about writing a clear but yet concise code base that you can pick up:

1. the function call in the docstring already states `Y::Dataset`, thus no reason to repeat this in the main body.
2. A comment that simply reads what the following piece of code explicitly says anyways, like
   ```
   # tree for input data
   vtree = KDTree(Y[1:NN], metric)
   ```
   isn't that useful. The code already conveys clearly enough what is happening. Great comments are those that bring critical new information regarding the algorithm or the implementation, like e.g. `# Eq. 16`.
3. Just using `vec(β)` is enough to make it into a vector, since the data were in the form of a "transposed" vector already.
4. Limiting `τs::AbstractRange` is not necessary for the algorithm. Using `τs = [1, 5, 15]` is still conceptually valid.

And my final comment is about the default values of `τs` which is `0:50`. Is there any point in applying the algorithm for `τ = 0` ? Perhaps this targets multi-variate input datasets I assume?

regarding this comment:

> Seems to me that we can finish the function MDOP by giving the total cycle number, and total τ range to search, and then from each curve obtain the global maximum: first obtain all local maxima, and then pick the highest of them. And then move to second cycle. As in your first figure.

Can we do this here, to leave for later only methods where the "selection of optimal parameters" is not clear? From what I've seen from your plots and also from the paper plots, here selecting the "optimal" βs is a simple call of `findmax()` on the β statistic. And then we can have a simple function that loops over some given number of cycles and just does the embedding?  Or maybe I have misunderstood the algorithm?

hkraemer:
> Just using vec(β) is enough to make it into a vector, since the data were in the form of a "transposed" vector already.

- Too easy :), cheers

> And my final comment is about the default values of τs which is 0:50. Is there any point in applying the algorithm for τ = 0 ? Perhaps this targets multi-variate input datasets I assume?

- That's right. for multivariate input this is (strictly) necessary.

> Can we do this here, to leave for later only methods where the "selection of optimal parameters" is not clear? From what I've seen from your plots and also from the paper plots, here selecting the "optimal" βs is a simple call of findmax() on the β statistic. And then we can have a simple function that loops over some given number of cycles and just does the embedding? Or maybe I have misunderstood the algorithm?

- I have implemented this in Matlab already. Basically it is as easy as you said. But the algorithm should automatically decide how many embedding cycles are needed. According to the paper the break criterion is a sufficiently small amount of false nearest neighbours (this is then a free parameter). Specifically, in the first cycle and for a multivariate input the algorithm needs to scan all possible combinations of timeseries as potential additional embedding components, i.e. one "needs to know with which (unlagged) time series to start with and which (lagged) time series to use as the second component of the embedding vector.  For all further embedding cycles one only needs to scan as many time as there are input time series, because one already has an embedding vector. I would also include this routine, called MDOP, in the MDOP.jl. But for now I think it is already okay to merge just the core algorithm (`beta_statistic`).

PR #44, opened 2020-07-28
Uzal cost

hkraemer:
I added another function `uzal_cost_local(...)`, in order to receive the L-statistic locally for each phase space point, as can be seen in the [paper's Fig. 16](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.84.016223) for instance. This might be of further use, since one can localize areas in the reconstruction space, which are "not properly unfolded yet".


src/uzal_cost.jl
```diff
@@ -115,6 +119,112 @@ function uzal_cost(Y::Dataset ; Tw::Int = 40, K::Int = 3, w::Int = 1, samplesize
...
+neighborhood. In contrast to the function `uzal_cost(...)`, `σ²` here does not
+get averaged over all the phase space reference points on the attractor.
+
+The `α`-statistic is a normalization factor, such that `σ`'s from different
+reconstructions can be compared. `α²` is defined as the inverse of the sum of
+the inverse of all `ϵ²`'s for all considered reference points.
+
+[^Uzal2011]: Uzal, L. C., Grinblat, G. L., Verdes, P. F. (2011). [Optimal reconstruction of dynamical systems: A noise amplification approach. Physical Review E 84, 016223](https://doi.org/10.1103/PhysRevE.84.016223).
+"""
+
```
Datseris, line 173: If there is an empty line between the docstring and the function, the docstring is not processed as attached to the function.

src/uzal_cost.jl
```diff
@@ -115,6 +119,112 @@ function uzal_cost(Y::Dataset ; Tw::Int = 40, K::Int = 3, w::Int = 1, samplesize
     return L
 end
 
+"""
+    uzal_cost_local(Y::Dataset; kwargs...) → L_local
```
Datseris, line 123: can't we just point to the documentation of `uzal_cost`, and only say here what is different with respect to `uzal_cost`, instead of repeating the whole docstring?

src/uzal_cost.jl
hkraemer: very useful information, thanks!

src/uzal_cost.jl
hkraemer: Of course

hkraemer: Is there a proper way of "pointing"?

Datseris: Yes! You can use this syntax: 
```
[`uzal_cost`](@ref)
```
which will create a corssreferenced link automatically to the existing docstrign.

hkraemer: Will do!

Datseris:
So far looks good. Do we need the file `testing_stuff.jl`? Also, purely for long-term stability, can you write a couple rudimentary tests that explicitly call `uzal_cost_local`? Then we can merge!

Datseris:
Before merging we should try to produce a figure like 2 or 16 of the Uzal paper, just to be sure everything makes sense. I'll use the same code that produces that figure in the documentation of DynamicalSystems.jl

hkraemer:
I delete `testing_stuff.jl` and will add two more tests explicitly calling `uzal_cost_local`. Where do you want me to produce the figures? Also in uzal_cost_test.jl?

Datseris:
Yeap, but push the plotting code commented!

hkraemer:
Alright. Done. I am quite confident that this is implemented correctly.
Local_cost_roessler

Datseris:
@hkraemer I've tried to merge this branch with `master` because there were some conflicts. At the moment it seems that there are some problems in the test suite. I get:

```
  Expression: maximum(L_local) > L
  MethodError: no method matching isless(::Array{Float64,2}, ::Float64)
```
it seems that you are comparing a number with `L`, which is an array (the Uzal cost for different parameters).

Can you please have a final look at the tests, make sure they pass locally, and then we can merge?

hkraemer:
so sorry for that, some little code block got missing. It works now

PR #45, opened 2020-08-10
Nichkawde (MDOP) method univariate embedding

hkraemer:
The automatized MDOP-embedding procedure from Nichkawde for univariate time series input.


hkraemer:
There is an issue with type-stability for the functions `estimate_maximum_delay()` and `MDOP()`. For illustration I used the `@code_warning`-macro from BenchmarkTools.jl and integrated that in the test_MDOP.jl-file. If I run that on the core-function `beta_statistic()` (line 52, 81) it is all good and the code is labeled type-stable. However, when I run that on the above mentioned functions (lines 105, 106, 141) I'll get a "red" warnsign, telling me there is an issue with the return types of the functions. So there might be an issue, due to the multiple return values, but I can not figure it out. When testing `estimate_maximum_delay()` for instance (lines 105, 106) I'll get the following output from `@code_warning`:

```
Variables
  #unused#::Core.Compiler.Const(DelayEmbeddings.var"#estimate_maximum_delay##kw"(), false)
  @_2::NamedTuple{(:tw, :samplesize),Tuple{UnitRange{Int64},Float64}}
  @_3::Core.Compiler.Const(DelayEmbeddings.estimate_maximum_delay, false)
  s::Dataset{1,Float64}
  tw::UnitRange{Int64}
  samplesize::Float64
  @_7::UnitRange{Int64}
  @_8::Float64

Body::Tuple{Int64,Dataset{_A,Float64} where _A}
1 ─ %1  = Base.haskey(@_2, :tw)::Core.Compiler.Const(true, false)
│         %1
│         (@_7 = Base.getindex(@_2, :tw))
└──       goto #3
2 ─       Core.Compiler.Const(:(@_7 = 1:50), false)
3 ┄       (tw = @_7)
│   %7  = Base.haskey(@_2, :samplesize)::Core.Compiler.Const(true, false)
│         %7
│   %9  = Base.getindex(@_2, :samplesize)::Float64
│   %10 = (%9 isa DelayEmbeddings.Float64)::Core.Compiler.Const(true, false)
│         %10
└──       goto #5
4 ─       Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :samplesize, DelayEmbeddings.Float64, %9)), false)
└──       Core.Compiler.Const(:(Core.throw(%13)), false)
5 ┄       (@_8 = %9)
└──       goto #7
6 ─       Core.Compiler.Const(:(@_8 = 1.0), false)
7 ┄       (samplesize = @_8)
│   %19 = (:tw, :samplesize)::Core.Compiler.Const((:tw, :samplesize), false)
│   %20 = Core.apply_type(Core.NamedTuple, %19)::Core.Compiler.Const(NamedTuple{(:tw, :samplesize),T} where T<:Tuple, false)
│   %21 = Base.structdiff(@_2, %20)::Core.Compiler.Const(NamedTuple(), false)
│   %22 = Base.pairs(%21)::Core.Compiler.Const(Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}(), false)
│   %23 = Base.isempty(%22)::Core.Compiler.Const(true, false)
│         %23
└──       goto #9
8 ─       Core.Compiler.Const(:(Base.kwerr(@_2, @_3, s)), false)
9 ┄ %27 = DelayEmbeddings.:(var"#estimate_maximum_delay#115")(tw, samplesize, @_3, s)::Tuple{Int64,Dataset{_A,Float64} where _A}
└──       return %27

```
Line 9 is complaining about the return (it is marked red). It knows it is a Tuple{Int64,Dataset{_A,Float64} parametrized on A. I don't know why this is not code-stable. Isn't A the type of the "dimensionality" of the Dataset? This should always be an `Int`. The according function reads

```
function estimate_maximum_delay(s::Vector{Float64}; tw=1:50, samplesize::Float64=1.0)
    L = zeros(Float64,length(tw),1)
    # loop over time windows
    cnt = 1
    for i in tw
        Y_act = embed(s,i,1)
        L[cnt] = uzal_cost(Y_act; samplesize = samplesize)
        cnt +=1
    end
    ~, τ_max = findmin(L)
    return tw[τ_max], Dataset(L)
end
```
So, in my view at least, I declared everything? 

A similar problem when testing `MDOP()` (line 141 in test_MDOP.jl). I'll get:

```
Variables
  #unused#::Core.Compiler.Const(DelayEmbeddings.var"#MDOP##kw"(), false)
  @_2::NamedTuple{(:τs, :w, :βs),Tuple{UnitRange{Int64},Int64,Bool}}
  @_3::Core.Compiler.Const(DelayEmbeddings.MDOP, false)
  s::Array{Float64,1}
  τs::UnitRange{Int64}
  w::Int64
  fnn_thres::Float64
  r::Float64
  βs::Bool
  @_10::UnitRange{Int64}
  @_11::Int64
  @_12::Float64
  @_13::Float64
  @_14::Bool

Body::Tuple
1 ── %1  = Base.haskey(@_2, :τs)::Core.Compiler.Const(true, false)
│          %1
│          (@_10 = Base.getindex(@_2, :τs))
└───       goto #3
2 ──       Core.Compiler.Const(:(@_10 = 0:50), false)
3 ┄─       (τs = @_10)
│    %7  = Base.haskey(@_2, :w)::Core.Compiler.Const(true, false)
│          %7
│    %9  = Base.getindex(@_2, :w)::Int64
│    %10 = (%9 isa DelayEmbeddings.Int)::Core.Compiler.Const(true, false)
│          %10
└───       goto #5
4 ──       Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :w, DelayEmbeddings.Int, %9)), false)
└───       Core.Compiler.Const(:(Core.throw(%13)), false)
5 ┄─       (@_11 = %9)
└───       goto #7
6 ──       Core.Compiler.Const(:(@_11 = 1), false)
7 ┄─       (w = @_11)
│    %19 = Base.haskey(@_2, :fnn_thres)::Core.Compiler.Const(false, false)
└───       goto #9 if not %19
8 ──       Core.Compiler.Const(:(Base.getindex(@_2, :fnn_thres)), false)
│          Core.Compiler.Const(:(%21 isa DelayEmbeddings.Float64), false)
│          Core.Compiler.Const(:(%22), false)
│          Core.Compiler.Const(:(goto %27), false)
│          Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :fnn_thres, DelayEmbeddings.Float64, %21)), false)
│          Core.Compiler.Const(:(Core.throw(%25)), false)
│          Core.Compiler.Const(:(@_12 = %21), false)
└───       Core.Compiler.Const(:(goto %30), false)
9 ┄─       (@_12 = 0.05)
│          (fnn_thres = @_12::Core.Compiler.Const(0.05, false))
│    %31 = Base.haskey(@_2, :r)::Core.Compiler.Const(false, false)
└───       goto #11 if not %31
10 ─       Core.Compiler.Const(:(Base.getindex(@_2, :r)), false)
│          Core.Compiler.Const(:(%33 isa DelayEmbeddings.Float64), false)
│          Core.Compiler.Const(:(%34), false)
│          Core.Compiler.Const(:(goto %39), false)
│          Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :r, DelayEmbeddings.Float64, %33)), false)
│          Core.Compiler.Const(:(Core.throw(%37)), false)
│          Core.Compiler.Const(:(@_13 = %33), false)
└───       Core.Compiler.Const(:(goto %42), false)
11 ┄       (@_13 = 2.0)
│          (r = @_13::Core.Compiler.Const(2.0, false))
│    %43 = Base.haskey(@_2, :βs)::Core.Compiler.Const(true, false)
│          %43
│    %45 = Base.getindex(@_2, :βs)::Bool
│    %46 = (%45 isa DelayEmbeddings.Bool)::Core.Compiler.Const(true, false)
│          %46
└───       goto #13
12 ─       Core.Compiler.Const(:(%new(Core.TypeError, Symbol("keyword argument"), :βs, DelayEmbeddings.Bool, %45)), false)
└───       Core.Compiler.Const(:(Core.throw(%49)), false)
13 ┄       (@_14 = %45)
└───       goto #15
14 ─       Core.Compiler.Const(:(@_14 = false), false)
15 ┄       (βs = @_14)
:τs, :w, :fnn_thres, :r, :βs)::Core.Compiler.Const((:τs, :w, :fnn_thres, :r, :βs), false)
│    %56 = Core.apply_type(Core.NamedTuple, %55)::Core.Compiler.Const(NamedTuple{(:τs, :w, :fnn_thres, :r, :βs),T} where T<:Tuple, false)
│    %57 = Base.structdiff(@_2, %56)::Core.Compiler.Const(NamedTuple(), false)
│    %58 = Base.pairs(%57)::Core.Compiler.Const(Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}(), false)
│    %59 = Base.isempty(%58)::Core.Compiler.Const(true, false)
│          %59
└───       goto #17
16 ─       Core.Compiler.Const(:(Base.kwerr(@_2, @_3, s)), false)
17 ┄ %63 = DelayEmbeddings.:(var"#MDOP#107")(τs, w, fnn_thres::Core.Compiler.Const(0.05, false), r::Core.Compiler.Const(2.0, false), βs, @_3, s)::Tuple
└───       return %63
```
 Here "::Tuple" in line 17 is marked in red. Same story somehow, but here it does not even know any of the type of the returned values?! 

Maybe I am simply not able to read the `@code_warning`-return messages properly. Maybe there is a problem, when assigning multiple return values in Julia I don't know about. I am grateful for any suggestions.

Datseris:
Great, thanks for the description! I'll have a look ASAP and let you know! I'll also do a review of the new methods!

hkraemer:
Sorry it must be `@code_warntype` instead of `@code_warning`, of course.

test/test_MDOP.jl
```diff
@@ -44,11 +41,14 @@ ss = zeros(length(s))
 [ss[i] = s[i][1] for i in 1:length(s)]
```
Datseris, line 41: `[ss[i] = s[i][1] for i in 1:length(s)]` is not a nice Julia syntax and should not be used for iteration, but only when you want to create a **new** array. Doing `[whatever for i in 1:N]` allocates a new array. In the current case, you should do instead:
```julia
for i in 1:length(s); ss[i] = s[i][1]; end
```
which is also one line.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+time series to choose from. If `βs=true` then the function also returns the
+`β`-statistic for each embedding cycle as a `List` of `Vector`s.
 
 [^Nichkawde2013]: Nichkawde, Chetan (2013). [Optimal state-space reconstruction using derivatives on projected manifold. Physical Review E 87, 022905](https://doi.org/10.1103/PhysRevE.87.022905).
+[^Hegger1999]: Hegger, Rainer and Kantz, Holger (1999). [Improved false nearest neighbor method to detect determinism in time series data. Physical Review E 60, 4970](https://doi.org/10.1103/PhysRevE.60.4970).
+[^Kennel1992]: Kennel, M. B., Brown, R., Abarbanel, H. D. I. (1992). [Determining embedding dimension for state-space reconstruction using a geometrical construction. Phys. Rev. A 45, 3403] (https://doi.org/10.1103/PhysRevA.45.3403).
 """
-function MDOP(Y...)
+function MDOP(s::Vector{Float64}; τs = 0:50 , w::Int = 1, fnn_thres::Float64 = 0.05,
+    r::Float64 = 2.0, βs::Bool=false)
```
Datseris, line 55: Notice that type asserting `Float64` is not a good idea: it makes your code not work with other numeric types. What you want to do instead is assert as follows:
```julia
s::Vector{<: Real} # vector of any real numbers
r::Real # any real number
```
This is **not** type unstable. It just allows more types while still annotating that you would expect a vector of numbers.

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
     return vec(β)
 end
+
+
+"""
+    embed2(Y::Dataset,s::Vector,τ::Int) -> Y_new
+adds the `τ` lagged values of the time series `s` as additional component to the
+embedding vector `Y`.
+"""
+function embed2(Y::Dataset,s::Vector{Float64},τ::Int)
```
Datseris, line 229: Please consider adding a white space after each comma: `Y, s, τ` instead of `Y,s,τ`, because it matches in style with the rest of the source code and thus makes it more readable :) 

(also, same comment as above for the assertion `s::Vector{Float64} -> s::Vector{<:Real}`.

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+        FNN = fnns/fnns2
+    end
+
+    return FNN
+end
+
+
+"""
+    estimate_maximum_delay(s::Vector; tw=1:50, samplesize::Float=1.0)) -> τ_max, L
+computes an upper bound for the search of optimal delays, when using `MDOP`
```
Datseris, line 314: Typically documentation strings in Julia are stated in imperative second person: "Compute an upper bound..." instead of "computes"

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+considered in the computation of `L` (see [`uzal_cost`](@ref)). When this
+statistic reaches its global minimum the maximum delay value `τ_max` gets
+returned. When `s` is a multivariate `Dataset`, `τ_max` will becomputed for all
+time series of that Dataset and the maximum value will be returned. The returned
+`L`-statistic is a Dataset of size (length(tw)*size(s,2)).
+
+[^Nichkawde2013]: Nichkawde, Chetan (2013). [Optimal state-space reconstruction using derivatives on projected manifold. Physical Review E 87, 022905](https://doi.org/10.1103/PhysRevE.87.022905).
+[^Uzal2011]: Uzal, L. C., Grinblat, G. L., Verdes, P. F. (2011). [Optimal reconstruction of dynamical systems: A noise amplification approach. Physical Review E 84, 016223](https://doi.org/10.1103/PhysRevE.84.016223).
+"""
+function estimate_maximum_delay(s::Vector{Float64}; tw=1:50, samplesize::Float64=1.0)
```
Datseris, line 331: Here you should instead define the function as:
```julia
estimate_maximum_delay(s::Vector{T}; tw=1:50, samplesize::Real=1.0) where {T<:Real}
```
and you can use `T` in the rest of the function. Specifically, you must then do `L = zeros(T,length(tw),1)` (although the definition of `L` will change completely, see next comment).

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+    # loop over time windows
+    cnt = 1
+    for i in tw
+        Y_act = embed(s,i,1)
+        L[cnt] = uzal_cost(Y_act; samplesize = samplesize)
+        cnt +=1
+    end
+    ~, τ_max = findmin(L)
+    return tw[τ_max], Dataset(L)
+end
```
Datseris, line 342: I now see the problem. Yes, here the creation `Dataset(L)` is indeed type unstable. Why? Well, because the **size** of a `Matrix` in Julia is not known by its type. a `Matrix{Float64}` can be 10 by 10 or it can be 2348 by 217 and still have the same type.

But more importantly, here you should not make `L` a matrix. It is just a vector, since its second dimension has size one. Why not just do `L = zeros(T, length(tw))`? Then doing `Dataset(L)` works just as well and is type stable. You can check this by doing:
```julia
@code_warntype Dataset(rand(100))
```

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+
+function estimate_maximum_delay(s::Dataset; tw=1:50, samplesize::Float64=1.0)
+    τs_max = zeros(Int,size(s,2))
+    Ls = zeros(Float64,length(tw),size(s,2))
```
Datseris, line 346: Here the situation is similar as above. First, don't use `Float64` in Ls, and use instead `eltype(s)`. This ensures type stability if a user uses Float32 type numbers as well. Next, again, because you make `Ls` a matrix, its size is not type-inferable. The size of a Dataset instead is type-inferrable. I'd recommend to change the function definition to:
```julia
estimate_maximum_delay(s::Dataset{D, T}; tw=1:50, samplesize::Real=1.0) where {D, T}
```
then, inside the source of the function you can use both `D` (the dataset dimensionality, i.e. `size(s, 2)`) and `T` (the dataset number type), when making `Ls`.

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+    L = zeros(Float64,length(tw),1)
+    # loop over time windows
+    cnt = 1
+    for i in tw
+        Y_act = embed(s,i,1)
+        L[cnt] = uzal_cost(Y_act; samplesize = samplesize)
+        cnt +=1
+    end
+    ~, τ_max = findmin(L)
+    return tw[τ_max], Dataset(L)
```
Datseris, line 341: here `L` is always a vector. Why do you make it into a `Dataset`? `Vector`s are meant to be kept `Vector`s, typically making a dataset with 1 dimension is not that useful.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+function MDOP(s::Vector{Float64}; τs = 0:50 , w::Int = 1, fnn_thres::Float64 = 0.05,
+    r::Float64 = 2.0, βs::Bool=false)
+    @assert 0 < fnn_thres < 1 "Please select a valid breaking criterion, i.e. a threshold value `fnn_thres` ∈ (0 1)"
+    @assert all(x -> x ≥ 0, τs)
+
+    max_num_of_cycles = 50 # assure that the algorithm will break after 50 embedding cycles
+
+    metric = Euclidean()
+    # normalize input time series (especially important for fnn-computation)
+    s = (s.-mean(s))./std(s)
```
Datseris, line 63: `s = (s.-mean(s))./std(s)` notice that this allocates unecessarily one extra copy of `s`. You want to do instead `s .= (s.-mean(s))./std(s)` (notice the dot before the `=`).

We already have a function `regularize(s::Dataset)`. Why don't you create the following:
```julia
regularize(s::Vector) = regularize!(copy(s))
regularize!(s::Vector) = (s .= (s.-mean(s))./std(s))
```
and we can re-use this function in the package?

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+    allNNidxs, NNdist_old = all_neighbors(vtree, Y_act, 1:length(Y_act), 1, w)
+
+    # set a flag, in order to tell the while loop when to stop. Each loop
+    # stands for encountering a new embedding dimension
+    flag = true
+
+    # set index-counter for the while loop
+    cnt = 1;
+
+    # preallocate output variables
```
Datseris, line 79: this comment is not necessary, because it is already clear from the source code what is happening.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+    τ_vals = zeros(Int,1)
+    ts_vals = ones(Int,1)
+    FNNs = zeros(Float64,1)
+    βs ? βS = Array{Float64, 2}(undef, length(τs), max_num_of_cycles) : nothing
```
Datseris, line 83: same comment regarding `Float64`: use generic type `T = eltype(s)` instead. Otherwise you will not have type stability if user uses other numeric types.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+
+
+    # loop over increasing embedding dimensions until some break criterion will
+    # tell the loop to stop/break
+    while flag
+
+        # get the β-statistic
+        β = beta_statistic(Y_act, s; τs = τs , w = w)
+        βs ? βS[:,cnt] = β : nothing
```
Datseris, line 92: I don't understand what you are trying to do here: `βs ? βS[:,cnt] = β : nothing` ? It is untypical in Julia to do assignment in such a manner. (same with doing `[s[i] = 3 for i in 1:length(s)]`. It is unintuitive syntax, because above the `nothing` is not used anywhere, while in the second example the list comprehension `[ ]` is also not used, since you don't use the created array.

Do instead:
```julia
if βs
  βS[:, cnt] .= β
end
```

I also have another request: `βs` and `βS` have variable names that appear very similar in code, yet their nature is completely different: one is a boolean, the other is a matrix. How about naming the boolean one something like `β_bool` or `β_condition`, or in general something that makes sense with respect to the algorithm?

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+
+        # determine the optimal tau value from the β-statistic
+        maxi, max_idx = findmax(β)
+        # store chosen delay (and chosen time series)
+        push!(τ_vals,τs[max_idx])
```
Datseris, line 97: If possible, please use white space after `,` in function arguments.

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+    # preallocation
+    Y_new = zeros(M,2)
+    # fill vector up until index M
+    Y_new[:,1] = y[1:M]
+    # add lagged component of s
+    Y_new[:,2] = s[1+τ:N]
+
+    return Dataset(Y_new)
+
+end
```
Datseris, line 265: There is a simple way to make this function and it is not only more efficient, but also much more concise:
```julia
function embed2(y, s, τ)
    N = length(y)
    @assert N == length(s)
    M = N - τ
    return Dataset(view(y, 1:M), view(s, τ+1:N))
end
```

Some helpful tips: in the line `Y_new[:,1] = y[1:M]`, the command `y[1:M]` creates a **copy** of `y`. You should instead use the `view` function. 
Then, you can directly create a `Dataset` from input vectors: `Dataset(x, y, z, s, ...)` there is no reason to first make a matrix.

Datseris: In addition, I find the name `embed2` unintuitive: it doesn't seem to indicate what the function does, and it is also too similar with the existing `embed` and thus can be confused... Can we try to come up with a better name?

Datseris: (also notice that the comments in the original function are not helpful because it is clear from the source code what is happening. Comments should *add* something to make the source code clearer. Otherwise, they only add an additional line of code and make the source file larger ;) )

src/MDOP.jl
```diff
@@ -87,3 +220,134 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+
+    # store fraction of valid nearest neighbors
+    if fnns2 == 0
+        FNN = NaN;
+    else
+        FNN = fnns/fnns2
+    end
+
+    return FNN
+end
```
Datseris, line 309: In the file `estimate_dimension.jl` we have a function called `fnn`. 

1. The two functions cannot have the same name because they overwrite each other.
2. Can you please give some comments on your method versus the existing one? If yous is better we should use it instead in general. 
3. Can you comment on the **performance** (computing time) of the two functions?

src/MDOP.jl
```diff
@@ -45,16 +45,15 @@ FNN-statistic `FNNs` increases . The final embedding vector is stored in `Y`
 `τ_vals` and the according time series number chosen for the according delay
 value in `τ_vals` is stored in `ts_vals`. For univariate embedding (`s::Vector`)
 `ts_vals` is a vector of ones of length `τ_vals`, because there is simply just
-time series to choose from. If `βs=true` then the function also returns the
-`β`-statistic for each embedding cycle as a `List` of `Vector`s.
+one `β`-statistic for each embedding cycle as a `List` of `Vector`s.
```
Datseris, line 48: `List` does not exist in Julia.

src/MDOP.jl
hkraemer: Thanks for that! At least my assumption about the `@code_warntype` output was kind of right :)
Btw. very helpful with explaining the parametrization of types!

src/MDOP.jl
hkraemer: jepp, makes sense. Will do

src/MDOP.jl
hkraemer: accepted

src/MDOP.jl
hkraemer: > 1. The two functions cannot have the same name because they overwrite each other.

Okay, I'll fix this. Is there any handy command, which tells you whether there is already a function in that package, which uses the same name?

> 2. Can you please give some comments on your method versus the existing one? If yous is better we should use it instead in general.

The existing methods in `estimate_dimension.jl` perform a false nearest neighbor for a given time series, using uniform time delay embedding and a predefined range of encountered dimensions. Here we need something else. We need just the basic idea of the FNN-method, namely counting the number of nearest neighbour distances, which exceed a certain threshold, when passing from one dimension into the next one. 
> when passing from one dimension into the next one.

in this context means to perform an embedding cycle with a delay, which is not known a priori. Consequently, the input to "my fnn"-function, are `Arrays` of nearest neighbour distances for dimension _**d**_ and for dimension _**d+1**_. That's it. Note that there is no nearest neighbour finding computations whatshowever. 

> 3. Can you comment on the performance (computing time) of the two functions?

As seen from my comments above, it does not make sense to compare a "full method" to my "in place, core" method.
If you want, we could draft "my fnn inplace function" to `estimate_dimension.jl`, if you want. It is handy, anyway. Furthermore it uses the idea of [Hegger & Kantz (1999)](https://doi.org/10.1103/PhysRevE.60.4970), whereas your functions `fnn` and `f1nn` use the original method of [Kennel et al. (1992)](https://journals.aps.org/pra/abstract/10.1103/PhysRevA.45.3403) (not so recommended in my opinion) and the method proposed by [Krakovska et al. (2015)](https://doi.org/10.1155/2015/932750), which I also not like that much.

Datseris: > Is there any handy command, which tells you whether there is already a function in that package, which uses the same name?

Yes, `isdefined(Module, :function_name)`.

---

For the remaining comments: okay, I see now. Sure, if you believe you have the time to implement a "full" version, feel free to do it. I also do not like the Krakovska method (in fact it is outperformed by methods published decades earlier).
For now you can focus on finishing this PR though (taking baby steps), and simply renaming `fnn` to something else, more fitting.

src/MDOP.jl
hkraemer: The `view`-function is very handy, indeed!

hkraemer: It seems, however, it is not possible to "slice" bigger pieces from a `Dataset`. I'll get error-messages on this one, when `y` is a Dataset:
```julia
function embed2(y, s, τ)
    N = length(y)
    @assert N == length(s)
    M = N - τ
    return Dataset(view(y, 1:M), view(s, τ+1:N))
end
```
> MethodError: no method matching view(::Dataset{1,Float64}, ::UnitRange{Int64}, ::Colon)
Closest candidates are:
  view(!Matched::StructArrays.StructArray{T,N,C,I} where I, ::Any...) where {T, N, C} at /Users/hkraemer/.julia/packages/StructArrays/OtfvU/src/structarray.jl:154
  view(!Matched::AbstractArray, ::Any...) where N at subarray.jl:155
  view(!Matched::AbstractUnitRange, ::AbstractUnitRange{var"#s91"} where var"#s91"<:Integer) at subarray.jl:167
  ...
embed2(::Dataset{1,Float64}, ::Array{Float64,1}, ::Int64) at MDOP.jl:235
MDOP(::Array{Float64,1}; τs::UnitRange{Int64}, w::Int64, fnn_thres::Float64, r::Int64, β_condition::Bool) at MDOP.jl:102
(::DelayEmbeddings.var"#MDOP##kw")(::NamedTuple{(:τs, :w, :β_condition),Tuple{UnitRange{Int64},Int64,Bool}}, ::typeof(MDOP), ::Array{Float64,1}) at MDOP.jl:56
top-level scope at test_MDOP.jl:142
include_string(::Function, ::Module, ::String, ::String) at loading.jl:1088

I could do:
```julia
function embed2(Y::Dataset{D,T}, s::Vector{T}, τ::Int) where {D, T<:Real}
    N = length(Y)
    @assert N == length(s)
    M = N - τ
    return Dataset(view(Y[:], 1:M), view(s, τ+1:N))
end
```
, but then `Dataset`-function does not know what to do. Finally,

```julia
function embed2(Y::Dataset{D,T}, s::Vector{T}, τ::Int) where {D, T<:Real}
    N = length(Y)
    @assert N <= length(s)
    M = N - τ
    return Dataset(hcat(view(Matrix(Y), 1:M, :), view(s, τ+1:N)))
end
```
works.

Datseris: Can you tell me **what** you want to do if `y` is a dataset? You want to get the values of which column at indices `1:M`? The last column? The first?

Datseris: And is the output of `embed2` always a dataset with 2 columns?  What does `embed2` do in general? If you describe what you want the function to do, perhaps we can find a simpler answer.

Datseris: Doing `Matrix(Y)` is bad: it copies the entire dataset into a matrix, and then copies that back into a dataset.

hkraemer: As described in the `docstring`
```julia
"""
    embed_one_cycle(Y, s::Vector, τ::Int) -> Y_new
Add the `τ` lagged values of the time series `s` as additional component to the
time series `Y` (`Vector` or `Dataset`), in order to form a higher embedded
vector `Y_new`. The dimensionality of `Y_new` , thus, equals the dimensionality
of `Y+1`.
"""
```
this function just adds another `τ` lagged version of the input `s`, as an additional component to `Y`.

> Doing Matrix(Y) is bad: it copies the entire dataset into a matrix, and then copies that back into a dataset.

Hm. I tried a lot, but `view` does not work that well on `Dataset`'s I have the feeling...

Datseris: Yes, `view` doesn't work with `Dataset` in the way you would expect, because it is a vector of static vectors internally. It just "behaves" like a matrix.

Gimme some time and I'll write a performant version of `embed_one_cycle` and then tag you in the commit so you can see what I did.

src/MDOP.jl
```diff
@@ -87,3 +220,70 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+function estimate_maximum_delay(s::Dataset{D,T}; tw=1:50, samplesize::Real=1) where {D, T<:Real}
+    M = size(s,2)
+    τs_max = zeros(Int,M)
+    Ls = zeros(T,length(tw),M)
+    for i = 1:M
+        τs_max[i], L = estimate_maximum_delay(vec(s[:,i]); tw = tw, samplesize = samplesize)
+        Ls[:,i] = Matrix(L)
+    end
+    return maximum(τs_max), Dataset(Ls)
+end
```
hkraemer: I am still unable to make this type-stable. Instead of `Ls = zeros(T,length(tw),M)` I would somehow need to preallocate a `Dataset`. This way, it does not work :(

Datseris: I'll have a look at everything again

src/MDOP.jl
hkraemer: :+1:

hkraemer:
- [x] Most functions now type-stable
- [ ] type-stability of `MDOP` and `estimate_maximum_delay(Y::Dataset...)`
- [ ] making `embed_one_cycle` more performant.

Thanks @Datseris for working yourself through that!

Datseris:
@hkraemer have a look at my latest commit. In general always keep the following in the back of your head: you should never intermediatelly transform your data into a matrix before making them a dataset. A dataset *behaves* like a matrix, but for performance reasons, it is something entirely different under the hood. A matrix can never give a type-stable dataset.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
 
 export beta_statistic
 export MDOP
+export estimate_maximum_delay
 
 
 """
-    MDOP(....)
-method by [^Nichkawde2013]
+    MDOP(s::Vector; kwargs...) → Y, τ_vals, ts_vals, FNNs [,βS]
```
Datseris, line 11: What does `MDOP` stand for?

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+* `w::Int = 1`: Theiler window (neighbors in time with index `w` close to the point,
+  that are excluded from being true neighbors). `w=0` means to exclude only the
+  point itself, and no temporal neighbors.
+* `fnn_thres::Real= 0.05`: A threshold value defining a sufficiently small fraction
+  of false nearest neighbors, in order to the let algorithm terminate and stop
+  the embedding procedure (`0 <= fnn_thres < 1).
+* `r::Real = 2`: The threshold for the tolerable relative increase of the distance
+  between the nearest neighbors, when increasing the embedding dimension.
+* `β_condition::Bool = false`: When set `true`, the function also returnes the `β`-statistics
+  of all embedding cycles.
```
Datseris, line 29: I don't think this keyword is necessary, because `βs` are computed anyways. The function already has at least 3 return values, let's just always return `βs` as well. It is uncommon that the **value** of an argument decides the **number** of returned elements. (it is also type unstable).

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+one `β`-statistic for each embedding cycle as an `Array` of `Vector`s.
 
 [^Nichkawde2013]: Nichkawde, Chetan (2013). [Optimal state-space reconstruction using derivatives on projected manifold. Physical Review E 87, 022905](https://doi.org/10.1103/PhysRevE.87.022905).
+[^Hegger1999]: Hegger, Rainer and Kantz, Holger (1999). [Improved false nearest neighbor method to detect determinism in time series data. Physical Review E 60, 4970](https://doi.org/10.1103/PhysRevE.60.4970).
+[^Kennel1992]: Kennel, M. B., Brown, R., Abarbanel, H. D. I. (1992). [Determining embedding dimension for state-space reconstruction using a geometrical construction. Phys. Rev. A 45, 3403] (https://doi.org/10.1103/PhysRevA.45.3403).
 """
-function MDOP(Y...)
+function MDOP(s::Vector{<:Real}; τs = 0:50 , w::Int = 1, fnn_thres::Real = 0.05,
+    r::Real = 2, β_condition::Bool=false)
+    @assert 0 <= fnn_thres < 1 "Please select a valid breaking criterion, i.e. a threshold value `fnn_thres` ∈ [0 1)"
```
Datseris, line 56: In Julia you can type `\le` and then TAB, and it will make `≤`. You should use this instead :)

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+    @assert all(x -> x ≥ 0, τs)
+
+    T = eltype(s)
+
+    max_num_of_cycles = 50 # assure that the algorithm will break after 50 embedding cycles
+
+    metric = Euclidean()
+    # normalize input time series (especially important for fnn-computation)
+    s .= (s.-mean(s))./std(s)
```
Datseris, line 65: Wait, here we are altering the input timeseries of the user, without warning. I Didn't realize this before hand when I recommended `.=`. 

Please do `s = regularize(s)` instead. This leaves the original timeseries unchanged. If not, it can make user code behave very weirdly without them knowing why.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+        if β_condition; βS[:, cnt] = β; end
+
+        # determine the optimal tau value from the β-statistic
+        maxi, max_idx = findmax(β)
+        # store chosen delay (and chosen time series)
+        push!(τ_vals, τs[max_idx])
+        push!(ts_vals, 1)
+
+        # create phase space vector for this embedding cycle
+        Y_act = DelayEmbeddings.embed_one_cycle(Y_act,s,τ_vals[cnt+1])
```
Datseris, line 102: Notice that this is a type-unstable operation. The dimensionality of a dataset is part of its type. Here you replace `Y_act` with a version of itself with different dimensionality and thus different type.

However, with this algorithm design, there isn't a better alterantive and we will have just ignore this type instability and focus on doing proper profiling instead to see if it affects anything.

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+
+        # compute nearest neighbor distances in the new embedding dimension for
+        # FNN statistic
```
Datseris, line 105: This comment doesn't add anything because the function name `all_neighbors` is clear enough in what it does.

src/estimate_dimension.jl
```diff
@@ -226,6 +226,49 @@ function _compare_first_nn(s, γ::Int, τ::Int, Rγ::Dataset{D,T}, metric) where
...
+set to 2.
+[^Hegger1999]: Hegger, Rainer and Kantz, Holger (1999). [Improved false nearest neighbor method to detect determinism in time series data. Physical Review E 60, 4970](https://doi.org/10.1103/PhysRevE.60.4970).
+"""
+function fnn_embedding_cycle(NNdist::T, NNdistnew::T; r::Real=2) where {T}
+    @assert length(NNdist) == length(NNdistnew) "Both input vectors need to store the same number of distances."
+
+    # convert array of arrays into simple vectors, since we only look at K=1
+    NN_old = zeros(length(NNdist))
+    NN_new = zeros(length(NNdistnew))
+    for i = 1:length(NNdist)
```
Datseris, line 246: add `@inbounds` before the `for`

src/MDOP.jl
```diff
@@ -87,3 +220,73 @@ function beta_statistic(Y::Dataset, s::Vector; τs = 0:50 , w::Int = 1)
...
+function estimate_maximum_delay(s::Vector{T}; tw=1:50, samplesize::Real=1) where {T<:Real}
+    L = zeros(T,length(tw))
+    # loop over time windows
+    cnt = 1
+    for i in tw
+        Y_act = embed(s,i,1)
+        L[cnt] = uzal_cost(Y_act; samplesize = samplesize)
+        cnt +=1
+    end
+    ~, τ_max = findmin(L)
```
Datseris, line 279: What's happening here? Why do you use the function `~` in such a way?

src/MDOP.jl
```diff
@@ -4,16 +4,150 @@ using Distances
...
+Chetan Nichkawde [^Nichkawde2013].
+[`uzal_cost`](@ref)
+
+
+[^Nichkawde2013]: Nichkawde, Chetan (2013). [Optimal state-space reconstruction using derivatives on projected manifold. Physical Review E 87, 022905](https://doi.org/10.1103/PhysRevE.87.022905).
+"""
+function MDOP()
     #tbd
 
 end
```
Datseris, line 153: Do we need this code? Seems to be left-over that can be deleted.

Datseris:
Once you address the current review, I'll mention the technique of function barriers and how they can make the code faster but also more readable. Meanwhyile have a look at https://docs.julialang.org/en/v1/manual/performance-tips/#kernel-functions-1

src/MDOP.jl
hkraemer: This is a place marker for the multivariate embedding method. So far, we just considered univariate input. This is going to be a new PR then (same holds for #38 ).

src/MDOP.jl
hkraemer: I actually just want to supress the output, since I don't need it for any computations. That is still Matlab-syntax I suppose. Is there a similar, clean way in Julia?

src/MDOP.jl
hkraemer: haha, you are right. I'll explain the abbrevation

src/MDOP.jl
Datseris: Yes, use `_` instead.

Datseris:
@hkraemer I'll work on this PR a bit now. My main priority is splitting the `MDOP` function into smaller functions that are combined together. I'll do this for two reasons:

1. A function reaches maximum readability when it does only one thing, and is composed by 10-20 lines of code.
2. Because of the way Julia's compiling and dispatch, having small functions is better of performance because it doesn't allow type instabilities to propagate throughout the full algorithm (function barriers).

Then, I'll do some profiling and try to see if anything can be optimized!

p.s.: I'll rename `MDOP` to `mdop_embedding`, firstly to avoid capital letters which in Julia style are reserved for Type names, and also to make it more indicative that this function does an embedding. I actually think we should rename all such functions to end with `embedding`, e.g. `pecora_embedding`.

Datseris:
@hkraemer I have a question about this code block:
```julia
        maxi, max_idx = findmax(β)
        if length(max_idx) == 0
            flag = false
            println("Algorithm could not pick a delay value from β-statistic. Increase the considered delays in `τs`-input. NO valid embedding achieved")
            continue
        end
```

here `max_idx` will always be an integer, no matter what. In julia `length(n::Int) = 1` for any integer. Thus, the `if` statement you have written will never, ever, activate. What was it that you were actually trying to check here?

Datseris:
Hi @hkraemer I found out that most algorithm time is spent in the function `uzal_cost`, and thus this function needs to be optimized first. I've noticed that there is a big problem with this function. In the innermost loop we have (lines 88-95):
```julia
        # construct neighborhood for this fiducial point
        neighborhood = zeros(K+1,size(Y,2))
        neighborhood[1,:] = v  # the fiducial point is included in the neighborhood
        neighborhood[2:K+1,:] = vcat(transpose(Y[NNidxs])...)

        # estimate size of the neighborhood
        pd = pairwise(metric,neighborhood, dims = 1)
        ϵ²[i] = (2/(K*(K+1))) * sum(pd.^2)  # Eq. 16
```
why do we have to create a matrix every time? This is big performance and memory waste. Can you describe me what operations should be done here? Also notice that when you do `pd.^2` this allocates yet one more matrix. What do we need to do? We need to get the sum of the square of all inter-point distances between fiducial points, correct? Because I don't see a reason why this operation as a whole would allocate at all.

Datseris:
For `uzal_cost` of lorenz system:

Performance of commit 8be4b4c2e8755610f7c46ca37cc7ba70f605cb88

> 11.748 ms (231509 allocations: 24.65 MiB)

Performance of commit 

> 49.914 ms (1265666 allocations: 88.40 MiB) 0b63714763c5ea86bfbbdf133b9f49f43d2c66ec

Datseris:
Alright, that is enough for today. I've increased the performance of uzal cost by a full order of magnitude. It used to take 54 ms for Lorenz test, now it takes 5 ms. I'll let this rest for a couple of days so that @hkraemer you can see what I've done. In principle I've just tried to eliminate as many allocations as possible. I haven't checked your `FNN` code yet, maybe you want to double check it after all these commits.

Datseris:
Tests fail at Lorenz Fig 7: https://travis-ci.org/github/JuliaDynamics/DelayEmbeddings.jl/jobs/718628476#L420

hkraemer:
It works perfectly on my machine... uhmmm. 
Bildschirmfoto 2020-08-18 um 09 45 14

hkraemer:
Figure_1

hkraemer:
It should be fine! Can not figure out why Travis gets what it gets...

Datseris:
@hkraemer what version of Julia you are using? If you are using anything less than 1.5 please update to 1.5 and try to fix the tests again. For Julia 1.5 the random number generator was changed to a faster version, which as a result gives different numbers for the same seed.

hkraemer:
I am running on 1.5

hkraemer:
Maybe Travis is not :)

Datseris:
Travis is also on 1.5:

https://travis-ci.org/github/JuliaDynamics/DelayEmbeddings.jl/jobs/718628476#L167

The problem is that we simply cannot merge this with failing tests, because it will never allow us to test new functionality DelayEmbeddings.jl online. Tests will always fail and thus we won't have any indication of whether it is the *new* functionality that fails, or the *old* one.

hkraemer:
In this example there is not even a random seed used. I starred at this now for ages and I do not have any idea what is going on. It just workes on my machine. And also the other test are quite similar, and they pass and give the results I get on my machine. I think the only way to see what is going on is to display some stuff, in order to see where the error occurs. If you show me how to run the travis tests and get access to the log-files, I can do it on my own.

Datseris:
@hkraemer 

> `println("Algorithm stopped due to rising FNNs.")`

Is this scenario a VALID embedding reached, or INVALID/INSUFFICIENT?

Datseris:
@hkraemer 

Why do we return:

`Y_act[:,1:counter-1], τ_vals[1:counter-1], ts_vals[1:counter-1], FNNs[1:counter-1], βS[:,1:counter-1]` , especially for the `Y_act`? Why `1:counter-1` instead of getting the last column as well?

Datseris:
@hkraemer I've actually included the mdop test file into the `runtest.jl` file. So far, none of these tests were run on CI. You will notice that depending on the operating system, the method finds different dimension of embeddings: on Windows tests pass, on Linux the method embeds at one dimension higher, https://travis-ci.org/github/JuliaDynamics/DelayEmbeddings.jl/jobs/719102962#L394 

Is this method really so brittle, that just changing operating system can give different embedding? If so, we should mention this in our comparison (and perhaps in general think of doing some "stability" comparison).

hkraemer:
Hey @Datseris 

> 
> > println("Algorithm stopped due to rising FNNs.")
> 
> Is this scenario a VALID embedding reached, or INVALID/INSUFFICIENT?

In my opinion it is VALID, but this has not been explicitly addressed in the paper. There it is just stated that the FNN's should fall below a certain threshold or tend to 0 (same for the Garcia & Almeida paper). This is impossible, when there is any noise contamination. Since Nichkawde highlights the fact that his method eliminates the maximal amount of FNN's in each embedding cycle, a rising amount of FNN's is suited as a good break criterion I suppose.

hkraemer:
@Datseris,

> Why do we return:
> `Y_act[:,1:counter-1], τ_vals[1:counter-1], ts_vals[1:counter-1], FNNs[1:counter-1], βS[:,1:counter-1] `, especially for the `Y_act`? Why `1:counter-1` instead of getting the last column as well?

The last column/value represents the results from the last encountered embedding cycle. This last cycle is not necessary, when the break criterion, i.e. suffiently small amount of FNNs, is fullfilled. The returned embedding values, thus, need to be
`Y_act[:,1:counter-1], τ_vals[1:counter-1]` and `ts_vals[1:counter-1]`. For `FNNs, βS` I thought about returning also the very last value, in order to illustrate the last encountered cycle. One would need to explain that in the docstring though. Maybe this is more userfriendly, what do you think?

hkraemer:
@Datseris ,

> @hkraemer I've actually included the mdop test file into the runtest.jl file. So far, none of these tests were run on CI. You will notice that depending on the operating system, the method finds different dimension of embeddings: on Windows tests pass, on Linux the method embeds at one dimension higher, https://travis-ci.org/github/JuliaDynamics/DelayEmbeddings.jl/jobs/719102962#L394
> 
> Is this method really so brittle, that just changing operating system can give different embedding? If so, we should mention this in our comparison (and perhaps in general think of doing some "stability" comparison).

This sounds to be a bug somewhere. The break criterion is solely based on the amount of FNNs, in these cases. The amount of FNNs is a simple function of two input vectors of distances. And these distances should be computed in the same manner, independent from the used OS. The same argument holds for the `test_uzal_cost.jl`, where for some reason Travis computes way-off-values to test, whereas it does all the very similar tests before (also sorting some vector and taking the minimum) as expected. 

Is there a possibility that the solvers of `DifferentialEquations.jl` and `DelayDiffEq` work differently on different OS's? Maybe then the test would deal with different time series input?

Datseris:
Hi @hkraemer ,

I've asked the DiffEq people whether the delay diff eq solvers might give different results on different systems and the answer is:

> yes, they can be slightly different because of architectures, differences in the amount of SIMD, and SIMD changes the result. If the system is futhermore chaotic, O(1) differences are expected.

Given that pretty much everything we are treating is chaotic, we have to take this into account. I've relaxed the tests some more. If tests still fail, please take some time to play around with test limits and ensure tests pass online. Do not strictly test of some result is a specific value, but test that it is in a reasonable range of values. In general play around with it a bit and push things until we have passing tests :)

---

I've just pushed a very important commit now that separates `mdop_embedding` into two functions, one that generates the new `Y_act` and one that just does the isolated cycle.

This stops propagating the type instability of generating a dataset of different dimension, and also leads to readable code. Pro tip: try to limit your functions to 20-25 lines of code. Is a function is longer, separate it into simpler building blocks.

---

@hkraemer here is some exercise to help you write more performant Julia code. Inside `fnn_embedding_cycle` I see it is full of non-necessary allocations like:
```julia
    ratio2 = NN_new./NN_old

    cond1 = ratio2 .> r
    cond2 = NN_old .< ratio1

    fnns = sum(cond1.*cond2)
    fnns2 = sum(NN_old .< ratio1)
```
all of these operations, like `ratio .> r` generate intermediate arrays that have no value (because they are neither returned, or used later in the function). Instead, transform the entire thing into a `for` loop, and manually count `fnns, fnns2`.

Datseris:
I see that `beta_statistic` calculates `all_neighbors`, but this calculation has already been done for its input argument `Y`. Why do we have to re-calculate the neighbors instead of just passing the calculation along? Getting the nearest neighbors is the most costly operation in this function, it should be done only when absolutely necessary.

src/MDOP.jl
```diff
@@ -3,37 +3,154 @@ using StatsBase
...
+    # tell the loop to stop/break
+    while flag
+        Y_act, NNdist_new = mdop_embedding_cycle!(
+            Y_act, flag, s, τs, w, counter, βS, τ_vals, metric, r,
+            FNNs, fnn_thres, ts_vals, NNdist_old
+        )
+        flag = mdop_break_criterion(FNNs, counter, fnn_thres, max_num_of_cycles)
+        counter += 1
+        # for the next embedding cycle save the distances of NN for this dimension
+        NNdist_old = NNdist_new
```
hkraemer, line 90: > I see that beta_statistic calculates all_neighbors, but this calculation has already been done for its input argument Y. Why do we have to re-calculate the neighbors instead of just passing the calculation along? Getting the nearest neighbors is the most costly operation in this function, it should be done only when absolutely necessary.

That is just partly correct. Indeed `beta_statistic` computes all neighbours, and is has to. For the computations of FNN's in each embedding cycle `fnn_embedding_cycle` needs all NN-distances for the present and all NN-distances for the corresponding next embedding dimension. `beta_statistic` computes, thus, the "new" NN-distances in each cycle and these distances are then being used for the next cycle as the "old" NN-distances (see marked code), so it is all good. Just for the very first iteration cycle (==1st embedding cycle) the "old" distances do not exist so we compute them in line 69.

hkraemer:
Hi @Datseris ,

> Hi @hkraemer ,
> 
> I've asked the DiffEq people whether the delay diff eq solvers might give different results on different systems and the answer is:
> 
> > yes, they can be slightly different because of architectures, differences in the amount of SIMD, and SIMD changes the result. If the system is futhermore chaotic, O(1) differences are expected.
> 
> Given that pretty much everything we are treating is chaotic, we have to take this into account. I've relaxed the tests some more. If tests still fail, please take some time to play around with test limits and ensure tests pass online. Do not strictly test of some result is a specific value, but test that it is in a reasonable range of values. In general play around with it a bit and push things until we have passing tests :)

I know you do not really want predefined time series to get loaded in the tests, but here it seems to me that this would really make sense. What if we generate the time series we are looking at and simply save them as .csv or whatever and load them for the corresponding tests. If then the Linux-Travis-CI still fails, we face a very different problem. Moreover, we could stick to the more strict tests, which I would very much advise here. In the Uzal case, as well as in the MDOP case, there is good reasoning why I chose the "strict" tests, i.e. the small boundaries of tested values. If we'd relax these boundaries too far I am not convinced that these tests are actually testing the ability of the algorithm correctly. What do you think?

hkraemer:
> EDIT: This PR has had a lot of commits already. If you want to spend time on `fnn_embedding_cycle`, please do it in a new PR after we merge this one.

I'll do this, thanks.

hkraemer:
> I've just pushed a very important commit now that separates `mdop_embedding` into two functions, one that generates the new `Y_act` and one that just does the isolated cycle.
> 
> This stops propagating the type instability of generating a dataset of different dimension, and also leads to readable code. Pro tip: try to limit your functions to 20-25 lines of code. Is a function is longer, separate it into simpler building blocks.

This looks indeed very neat ;) I appreciate this

Datseris:
Hi @hkraemer finally tests pass. However we still won't upload the timeseries here. We will upload them to a separate repo and download them for tests. I anyways needed to upload some timeseries somewhere, as exercises for students, so I'll now kill two birds with one stone.

Can you please give me the **exact** code you used to generate each timeseries? Exact parmaeter combinations, etc.? By the way, the files that are just the timeseries of Lorenz/Roessler are not needed: one can read the trajectory file and keep the first column....

hkraemer:
Hey @Datseris,
that is good to hear. I left the code containing the exact settings to produce the time series in the file. It is outcommented just before the `readdlm`-statement. I would suggest to leave this outcommented code blocks in the files!

Datseris:
Alright, **tests are finally passing** so I'll be merging this one and tag a new version of DelayEmbeddings.jl with this PR in.

@hkraemer please update all other PRs so that they have passing tests. Because we've spent a lot of time here fixing the tests, for the remaining reviews, I'll only be looking at them if they have passing tests.

PR #46, opened 2020-08-13
Pecora method update

hkraemer:
I fine-tuned the Pecora method and set up a proper test file.


hkraemer:
It is now possible to input (more or less) important parameters for obtaining the continuity statistic, namely `α ` and `p`. Moreover, the amount of points for rejecting the Null now gets computed automatically and the continuity statistic is based on considering a range of `δ`-neighborhood-sizes (see docstring). Also the "hard-coded" input `N` of considered numbers of fiducial points is replaced by `samplesize`, which reflects a relative measure (a fraction of all points) and is, thus, more user-friendly. Finally I set up a proper test file.

test/test_pecora.jl
```diff
@@ -1,65 +1,108 @@
...
+# ax2 = subplot(212)
+# xlabel("τ (index units)")
+# ylabel("⟨Γ⟩")
+
 
 τs = (0,)
-@time es, Γs = pecora(s, τs; T = 1:Tmax, N = 100, metric = metric, undersampling = UNDERSAMPLING)
-ax1.plot(es, label = "τs = $(τs)")
-ax2.plot(Γs)
+@code_warntype pecora(s, τs; delays = 0:Tmax, w = optimal_τ, samplesize = samplesize, K = K, metric = metric, undersampling = UNDERSAMPLING)
```
hkraemer, line 40: Here I get a similar issue as in #45 and #38. I can not figure out why it is not typestable :/

test/test_pecora.jl
```diff
@@ -1,10 +1,8 @@
-using DelayEmbeddings, DynamicalSystemsBase, DelimitedFiles
+using DelayEmbeddings, DynamicalSystemsBase
 using DifferentialEquations
-using Distances
 using Random
 using Test
-using Peaks
-using BenchmarkTools
+import Peaks
```
Datseris, line 5: Peaks is not in the Project.toml file under [extras]

Project.toml
```diff
@@ -23,10 +24,12 @@ StatsBase = "0.24, 0.32, 0.33"
 julia = "1"
 
 [extras]
+DelayDiffEq = "bcd4f6db-9728-5f36-b5f7-82caef46ccdb"
```
Datseris, line 27: I've just checked the tests, but I didn't see `DelayDiffEq` used somewhere. Is it necessary? If you need a MackieGlass timeseries, we now have it uploaded (for the Nickhwade PR)

Datseris:
Great!

PR #47, opened 2020-08-24
Fixed bug in min_pairwise_distance and mutinfo

apbraun:
The function `knn` from `NearestNeighbors` has also been imported from `Neighborhood` it has not been explicitly stated from which package `knn` was imported in `mutinfo` and `min_pairwise_distance`. Just changed that.


Datseris:
Thanks @apbraun , this is a correct PR that correctly fixes the problem. The only reason I won't merge it is because we have already fixed this in PR #45 , which is an extremely long (unecessarily long) PR that I will merge today. Thus, I'll try to avoid merge conflicts and just let everything happen on #45.

PR #51, opened 2020-08-28
bug-fixed epsilon-star-statistic; bug-fixed temporal range neighbors

hkraemer:
I discovered another bug, while comparing results of `pecora`, specifcally `continuity_per_timeseries`, to the results of my Matlab implementation. I fixed them in a few lines of code. Furthermore I shifted the computation of the dictionary for the δ-Neighborhood (calling `get_binomial_table`) from `ε★` into `continuity_per_timeseries`. This way it gets computed just once, instead of giving the same results, when calling `ε★` with the same parameters.

### DIFF
--- a/src/pecora.jl
+++ b/src/pecora.jl
@@ -219,7 +219,7 @@ function pecora(
     end
     undersampling && metric ≠ Chebyshev() && error("Chebyshev metric required for undersampling")
     vspace = genembed(s, τs, js)
-    vtree = KDTree(vspace.data, metric)
+    vtree = KDTree(vspace.data[1:end-maximum(delays)], metric)
     all_ε★ = zeros(length(delays), length(J))
     all_Γ = copy(all_ε★)
     allts = columns(s)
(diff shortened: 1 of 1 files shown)


src/pecora.jl
```diff
@@ -219,7 +219,7 @@ function pecora(
     end
     undersampling && metric ≠ Chebyshev() && error("Chebyshev metric required for undersampling")
     vspace = genembed(s, τs, js)
-    vtree = KDTree(vspace.data, metric)
+    vtree = KDTree(vspace.data[1:end-maximum(delays)], metric)
```
hkraemer, line 222: Here I constrained the total number of points (including the nearest neighbours) with respect to the maximum considered delay time.

src/pecora.jl
```diff
@@ -263,14 +263,15 @@ columns(s::AbstractVector) = (s, )
...
     avrg_ε★ = zeros(size(delays))
     Ks = [k for k in 8:K]
+    δ_to_ε_amount = get_binomial_table(p, α; trial_range = length(Ks))
     for (ι, τ) in enumerate(delays) # Loop over the different delays
         c = 0
         for (i, n) in enumerate(ns) # Loop over fiducial points
             NNidxs = allNNidxs[i] # indices of k nearest neighbors to v
             # Check if any of the indices of the neighbors falls out of temporal range
-            any(j -> (j+τ > length(x)) | (j+τ < 1), NNidxs) && continue
+            any(j -> (j+τ < 1), NNidxs) && continue
```
hkraemer, line 272: This first check is not necessary anymore. See my first comment above.

src/pecora.jl
```diff
@@ -281,12 +282,11 @@ function continuity_per_timeseries(x::AbstractVector, ns, allNNidxs, delays, K,
...
 
-function ε★(x, n, τ, NNidxs, K::AbstractVector, α, p)
-    δ_to_ε_amount = get_binomial_table(p, α; trial_range = length(K))
+function ε★(x, n, τ, NNidxs, δ_to_ε_amount, Ks)
     a = x[n+τ] # fiducial point in ε-space
     @inbounds dis = [abs(a - x[i+τ]) for i in NNidxs]
-    ε = zeros(length(K))
-    for (i, k) in enumerate(K)
+    ε = zeros(length(Ks))
+    for (i, k) in enumerate(Ks)
```
hkraemer, line 289: adapted nomenclature

hkraemer:
For higher sample sizes we use in the tests, this bug didn't lead to a substantial different behavior. But when I encountered smaller time series inputs the resulting continuity statistics looked quite different. As expected, the unecessary former suppressing of neighbours plays a role, when the samplesize (number of fiducial points) is rather small. Now it's all good.

Datseris:
Great!!!

PR #52, opened 2020-08-31
Bug fix hcat lagged values

hkraemer:
There was a little bug in the function `hcat_lagged_values(Y::Vector, s::Vector, ...)`. It was concatenating just the values from `s`, not from `Y`. I also moved this into `various.jl`.


Datseris:
Ops, that was quite stupid of me! Thanks!~!!

PR #53, opened 2020-09-09
Multivariate embedding now possible; adjusted right output

hkraemer:
I completed the method by adding the possibilty for multivariate input. I also adjusted the final output reconstruction vector also for the univariate case: It is now not normalized (i.e. uses the same data, which has been inputed) and also has the right length. Before it was too short - according to the very last encountered embedding cycle. But the very last embedding cycle never plays a role in the final output, since the breaking criterion will determine the stopping. Now the length is in accordance with the output values for the taus and the corresponding time series.

### DIFF
--- a/src/MDOP.jl
+++ b/src/MDOP.jl
@@ -46,6 +46,9 @@ The final embedding is returned as `Y`. The chosen delay values for each embeddi
 cycle are stored in the `τ_vals` and the according timeseries index chosen for the
 the respective according delay value in `τ_vals` is stored in `ts_vals`.
 `βS, FNNs` are returned for clarity and double-checking, since they are computed anyway.
+In case of multivariate embedding, `βS` will store all `β`-statistics for all
+available time series in each embedding cycle. To double-check the actual used
+`β`-statistics in an embedding cycle 'k', simply `βS[k][:,ts_vals[k+1]]`.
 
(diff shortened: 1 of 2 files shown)


hkraemer:
Another comment on `mdop_multivariate_embedding_cycle!`. This is quite long, i.e. twice as long as the roughly suggested 20 lines of code. The reason is, that there is a case distinction whether or not one is in the first embedding cycle or not. Cleary, one could easily split up the two cases in another two subfunctions, in order to increase code-readibility for `mdop_multivariate_embedding_cycle!`. - But since there are so many subfunctions already, I had the feeling this would be even more complicated to understand. @Datseris , please tell me your opinion about this. It would not be a big deal anyway to split it up.

Datseris:
> But since there are so many subfunctions already, I had the feeling this would be even more complicated to understand. @Datseris , please tell me your opinion about this. It would not be a big deal anyway to split it up.

My opinion on this is set in stone: the simpler the function, the better. I try to design my functions so that they only have a singular role that can be described by their name. For me a typically good breaking point is when a function exceeds 50 loc or it exceeds two nesting levels (e.g. a while inside a function and then a for inside the while).

I also believe that more functions, each with small size, are easier to read than a larger function that contains the previous ones. A small function offers a focus point: I understand a specific, self-contained piece of code. Once I understand it, I only need the function's name to remember what it does.

That being said, these are only rough guidelines of course. In general I don't find the current PR too hard to read. I would argue it is useful for `mdop_multivariate_embedding_cycle!` being split into two, because it is a bit large, but this is entirely up to you to decide.

hkraemer:
jaja, you are right :-)

PR #54, opened 2020-09-10
Garcia multivariate

hkraemer:
As in #53 , this is the extension of #38 for multivariate data input. Again, because it is based on the same `N`-statistic, I could not reproduce the results shown in the [paper](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.72.027205), i.e. Fig. 2a., similar to the  trial of reproducing Fig.2 of the paper [paper](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.71.037204) in #38 . Until now, Prof. Almeida did not respond to my kind Email I sent him on June 25th. As already discussed in #38 I am quite confident that I implemented the N-statistic right, at least, the way they proposed it (crossvalidation with two other independent implementations by some other researchers of the same thing).

### DIFF
--- a/test/test_garcia.jl
+++ b/test/test_garcia.jl
@@ -12,9 +12,10 @@ tr = trajectory(lo, 60; dt = 0.01, Ttr = 10)
 
 x = tr[:, 1]
 Y = Dataset(x)
+Y = regularize(Y)
 
 @testset "N-statistic" begin
-    τs = 0:50
+    τs = 0:100
     N , NN_distances = n_statistic(Y, x, w=0, T=1, τs = τs)
 
     T = 17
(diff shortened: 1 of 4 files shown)


Datseris:
:P yeah, welcome to my world: a large percentage of the methods I've implemented in DynamicalSystems.jl were actually not reproducible with respect to the papers........

Anyways, I'll do a review here and merge

Datseris:
Hi Hauke, can you please also do:

1. In `Project.toml` increment minor version.
2. In `CHANGELOG.md` add relevant entries for your two recent PRs

Thanks!

test/test_garcia.jl
```diff
@@ -28,9 +29,12 @@ Y = Dataset(x)
     max_1_idx = Peaks.maxima(N,min_dist)
     max_2_idx = Peaks.maxima(N2,min_dist)
 
-    @test max_1_idx == max_2_idx
```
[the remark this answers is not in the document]
hkraemer: I had to adjust it, because I changed the time series length to the values described in the paper. - Just to see, that it still does not match the results proposed there.

test/test_garcia.jl
```diff
@@ -28,9 +29,12 @@ Y = Dataset(x)
...
+    @test 30 ≤ max_1_idx[1] ≤ 35
+    @test 30 ≤ max_2_idx[1] ≤ 35
+    @test 69 ≤ max_1_idx[2] ≤ 75
+    @test 69 ≤ max_2_idx[2] ≤ 75
 
-    # # plot N-Statistic for the Lorenz system as in Fig. 2(a) in [^Garcia2005b]
+    # # plot N-Statistic for the Lorenz system as in Fig. 2(a) in [^Garcia2005a]
```
Datseris, line 37: I wonder, should we add a comment here that our code fails to reproduce the paper? If someone runs this code and doesn't get the same figure they might be like "do I have an incorrect installation?"

hkraemer: Yeah, I also thought about it. How did you proceed in those cases for all the stuff you could not reproduce? Or maybe better to leave a note in the docs?!

Datseris: I write it down in both comments in the tests as well as in the docstrings. :)

Datseris: For example see my comment at the end of this docstring: https://juliadynamics.github.io/SignalDecomposition.jl/dev/#SignalDecomposition.ManifoldProjection . I am very explicit about what is wrong with the paper.

PR #57, opened 2020-10-03
hegger dimension and automated Cao's method

hkraemer:
Please have a look at `fnn_uniform_hegger`, which estimates embedding dimension according to a given uniform embedding. This could be easily extented by a corresponding `fnn_nonuniform_hegger`, where we use athe result of the `genembed`-function. I was not able to reproduce Fig. 2 in the paper.


Datseris:
(use file `test_fnn_hegger` to add tests)

Datseris:
@hkraemer alright, I did a bunch of re-ordering here and I think I have concluded on the final design.

1. At the moment we have methods like `afnn, fnn`, etc.. Hegger's method `fnn_uniform_hegger` should be named `ifnn` to follow the convention (from "improved" false nearest neighbors).
2. The other functions `afnn, fnn` do not return an embedding nor an embedding dimension. Instead they return the numerical value one can use to deduce the optimal dimension (i.e. Cao's method returns the ratio). I believe the function `fnn_uniform_hegger` should be doing the same thing and return the discriminatory statistic whatever this may be. The _new function_ `optimal_traditional_de` (which is the merge of `standard_embedding_hegger` and `standard_embedding_cao`) can return an embedding directly.
3. I am not familiar with Hegger's method, but what I think `ifnn` should return is the `FNNs`, correct?

hkraemer:
Yes, that is correct, it computes FNNs. The renaming totally makes sense.

hkraemer:
I setup a structure already you might want to change. It works already for all the different methods, but I realized that `fnn`-function gave a non-normalized value. So I normalized it, but now it looks like the FNN-fraction is way too small. I'll figure that out. I'll leave it for now, since you might have your own structure in mind here. Let me know when I can help.

hkraemer:
ahh. I see, now with your changes (deprecate `estimate_dimension`) it does not work right anymore. Still, you could use the structure with minor modifications. If you ask me, I would keep `estimate_dimension` and also `estimate_delay` with the corresponding method-keywords. Why not? These function nicely pipe all methods (and the ones which are about to be implemented), which, in turn, could then easily be called by any automatazion-function in a "clean" way.

hkraemer:
@Datseris, please keep in mind the necessary normalization of all FNN/dimension-estimator statistics (specifically, see `fnn`) and that the core of `ifnn` could not reproduce Fig. 2 of the corresponding paper yet.

Datseris:
@hkraemer I've noticed in some code parts you were writing things like:
```
a = estimate_dimension(s, τ, γs = 1:dmax, method = dimensionmethod)
```
however this is invalid syntax because for `estimate_dimension` both arguments are positional arguments and not keyword arguments. You'd have to write directly `estimate_dimension(s, τ, 1:dmax, dimensionmethod)` . Notice also that `γ` is the number of additional temporal entries in the delay vector, i.e. γ = dimension - 1.

Just giving a comment. For the code this won't matter because estimate_dimension will be removed completely.

Datseris:
@hkraemer please have a look at the current status, especially the function `optimal_traditional_de`. I'm happy with current design let me know what you think.

Something important that you have to check. The existing functions `afnn, fnn, f1nn` and `estimate_dimension` used the `reconstruct` function, which in contrast to embed, takes as an input `γ`, the _extra amount of delayed entries to add_. Thus, `γ = d-1` if `d` is the embedding dimension. I see you re-use the same `γ` in your new ifnn method, but you don't use `reconstruct`. Can you just make sure that the code you wrote is not messing up the "-1" factors and indeed finds correct dimensions?

In the meantime, I will code the function that checks local slopes until they become close to zero

Datseris:
Hi @hkraemer , please have a look at the file `test/compare_different_dimension_estimation.jl`. There I have some plotting code that compares the different dimensions and evaluates the result. I transform all of them in a format that is easy to compare, which means that FNN is normalized while Cao's result becomes 1 - itself. 

The condition I came up with for the estimation of the optimal embedding dimension was to check the earliest dimension that satisfies both the estimator less than a threshold r = 0.2, and the local slope less than a threshold z = 0.2. Please have a look, perhaps test it with the experimental system as well? 

Here are some example plots. I think the method is fine for a blind automated approach.

image


image

image


image

(duffing oscillator should have 2 embedding dimension even for time dependence, because default parameters lead to periodic orbit)

image

src/traditional_de/automated.jl
```diff
@@ -51,28 +52,29 @@ w::Int=1, rtol=10.0, atol=2.0, τs = 1:100, metric = Euclidean()
...
     τ = estimate_delay(s, delaymethod, τs)
-    γs = 1:dmax-1 # TODO: This must be updated to just dimension only
+    γs = 1:dmax-1 # TODO: This must be updated to dimension in 2.0
 
     if dimensionmethod=="afnn"
-        dimension_statistic = estimate_dimension(s, τ, γs, dimensionmethod)
+        dimension_statistic = afnn(s, τ, γs, metric)
         Y, τ = cao_embed(s, τ, dimension_statistic, thres)
         E2 = stochastic_indicator(s, τ, γs)
         flag = is_stochastic(E2, thres)
```
hkraemer, line 57: This one is still based on the assumption that E2 converges to 1, which I believe is not always the case. I will fix that by replacing it with the slope-convergence as you have suggested it.

src/traditional_de/automated.jl
```diff
@@ -51,28 +52,29 @@ w::Int=1, rtol=10.0, atol=2.0, τs = 1:100, metric = Euclidean()
...
         flag && println("Stochastic signal, valid embedding NOT achieved ⨉.")
-    elseif dimensionmethod=="ifnn"
-        dimension_statistic = estimate_dimension(s, τ, γs)
-        Y, τ = fnn_embed(s, τ, dimension_statistic, thres)
     elseif dimensionmethod=="fnn"
-        dimension_statistic = estimate_dimension(s, τ, γs, dimensionmethod; rtol = rtol, atol = atol)
+        dimension_statistic = fnn(s, τ, γs; rtol, atol)
```
hkraemer, line 60: We need to normalize the output!!

hkraemer: I'll do that asap. Then your normalization in `test/compare_different_dimension_estimation.jl` with respect to the maximum value. It needs to get normalized `fnn`-internally with respect to all encountered distances. I'll do that asap.

hkraemer:
Hey @Datseris ,
thanks for keep working on this. I am totally happy with the structure of `optimal_traditional_de`. However, regarding your examples and suggestions in `test/compare_different_dimension_estimation.jl` I would do it differently. For Cao's method I would only rely on the slopes, not another threshold for the vicinity of some absolute level, since this is what the statisic is all about. In contrast, the other FNN-methods do not necessarily depend on any slope as a breaking criterion. It is kind of implicitly incorporated in the algorithms, that is indeed true. - Since the statistics should stay at a minimum level with increasing embedding dimension, once the optimal embedding dimension is reached. But looking at noisy data, that might not be valid. The problems with all these methods is, except from Cao's approach, is that one has to adjust the fnn-threshold to the noiselevel of the data a priori. I think a Default-value of `fnn_thres=0.05`, as I suggested, is a good starting point. Long story short: For Cao's method we just focus on the slopes (for E1 and for the `is_stochastic`-function) and for all the other methods we focus on approaching the sufficiently small number of FNNs with respect to the set threshold AND the slope, as you suggested. I will implement all of this now and you can find the changes in the upcoming commit. Feel free to comment on this. ** While playing around with `test/compare_different_dimension_estimation.jl` for different setups/systems, I realized that the `f1nn` seems to fail everywhere. Even for noisefree data on deterministic data. Are you sure that is implemented right? Or is this method simply that bad?

hkraemer:
** I also corrected `ifnn` with respect to the notation of `γ = d-1`. Thanks for pointing that out @Datseris !

Datseris:
Hi @hkraemer okay, I agree with your plan. Are you working on implementing this or should I do it?

Datseris:
Regarding `f1nn`: hm, unfortunately I do not remember... When we first implemented it 2 years ago I thought "oh wow, this method sucks". But now that you mention it, perhaps we should have a second look whether the implementation is correct.... Its been a long time.

hkraemer:
I’ll do it first and will leave some questions for discussion and further work. I’ll push something in a minute.

hkraemer:
I incorporated the discussed changes in `optimal_traditional_de`-function. Simply run `test/compare_different_dimension_estimations.jl` and look at the results (see attached figures). Basically, it is doing what I'd expect it should do. However, the condition of a vanishing slope with respect to a `slope_thres` for the FNN-estimators can lead to overestimated embedding dimensions (see e.g. `fnn` for the Roessler or Lorenz system). If one lowers this threshold one receives "better" estimates, but at the same time Cao's (`afnn`) method will give wrong results, if the `E₁`-statistic increases slowly for small embedding dimensions (c.f. Lorenz96). One would simply need to choose a rather low `slope_thres` in case of `afnn` (I set the Default to 0.05, which seems to be quite a good choice) and a higher `slope_thres` for the other FNN-based estimators (e.g. 0.2 or so).
henon_c
henon_p
lorenz
lorenz96
roessler
sine
white_noise

hkraemer:
Note that `afnn` detects the white noise as a stochastic signal (this can not be seen in the figures, but when you run the mentioned file), but does not treat the AR(1)-process as a stochastic source. Moreover, whenever the FNN-statistic increases, the functions also throw the expected "double check statistic"-warning (e.g. for White noise). I think one could now deploy some tests on `test/compare_different_dimension_estimations.jl ` and also take care of the warnings thrown by the functions somehow.

src/traditional_de/automated.jl
```diff
@@ -28,10 +29,24 @@ For more details, see individual methods: [`afnn`](@ref), [`ifnn`](@ref),
...
+rtol=10.0, atol=2.0, τs = 1:100, metric = Euclidean(), r::Real=2.0
 ```
 
+## Description
+We estimate the optimal embedding dimension based on the given delay time gained
+from `dmethod` as follows: For Cao's method the optimal dimension is reached,
+when the slope of the `E₁`-statistic (output from `"afnn"`) falls below the
+threshold `slope_thres` (Default is .05) and the according stochastic test turns
+out to be false, i.e. when the `E₂`-statistic is not "equal" to 1 for all en-
+countered dimensions. The treat `E₂`-values as equal to 1, when `1-E₂ ≤ fnn_thres`.
```
Datseris, line 42:
```suggestion
countered dimensions. We treat `E₂`-values as equal to 1, when `1-E₂ ≤ fnn_thres`.
```

Datseris:
@hkraemer haha instead of copypasting everything and writing:
```
dict_keys = ["White noise","AR(1)","Hénon (chaotic)","Hénon (periodic)","Lorenz","Lorenz96","Roessler","Sine wave"]
```
you can just do
```
dict_keys = collect(keys(trajectories))
```

Datseris:
Hi Hauke, just went through the figures. I'm happy, are you also happy with the status? Then I put the code in the `optimala_traditional_de` function and we can merge? I will put a message to the documentation string that we provide an automated method but if one wants to be careful they should analyze the direct output instead of the automated choice.

hkraemer:
Jeah, I am also happy. We should, however, still take a deeper look into the `f1nn` implementation and also try to reproduce the figures in the Hegger-paper, concerning `ifnn`. But this can be done separately, no problem.

Datseris:
Okay then I'll write down some basic tests and merge this.

Datseris:
Hi @hkraemer 

Alright, this was a huge amount of work, but I'm finally done with it. In the coming DynamicalSystems.jl v2.0 I want to phase out all this weird `γs` thing that is a remnant of a very long past. I want to have only `embed` and to only use the embedding dimension `d`. The new `genembed` interface makes this `γs` nonsensical.

So, I'm taking steps to deprecate things out, and I started here. I've changed all source code to use `ds` instead of `γs`. In addition, I change all function names from `afnn, ifnn` to be `delay_afnn, delay_ifnn`. I needed a second name to make this deprecation non breaking, but I also thought that having tiny names might not be the best idea.

In anycase, I've changed the tests to use the Lorenz96 model instead of the Lorenz63. The only problem that remains for me is that the Cao's method algorithm fails to detect a correct embedding dimension for the Lorenz96 timeesries. This we have already seen in the last plots you have uploaded, which there you see Cao's method showin that the "correct" embedding dimension is 1. This is clearly wrong and we should fix it before merging. Do you have any idea why this happens?

Datseris:
This code shows all methods at the same time:

```julia
using DelayEmbeddings, DynamicalSystemsBase, PyPlot

ds = Systems.roessler()
# This trajectory is a chaotic attractor with fractal dim ≈ 2
# therefore the set needs at least embedding dimension of 3
lo = Systems.lorenz96(4, [0.1, 0.2, 0.5, 0.1]; F = 16.0);
data = trajectory(lo,5000.0;dt=0.05, Ttr = 100.0, diffeq...)
x = data[:,1]

dmax = 7
figure();
# for (i, delay_method) in enumerate(["afnn", "ifnn", "fnn", "f1nn"])
for (i, method) in enumerate(["afnn", "fnn", "f1nn", "ifnn"])
    # Plot statistic used to estimate optimal embedding
    # as well as the automated output embedding
    𝒟, τ, E = optimal_traditional_de(x, method; dmax)
    plot(1:dmax, E; label = method, marker = "o", ms = 5, color = "C$(i-1)")
    optimal_d = size(𝒟, 2)
    scatter(optimal_d, E[optimal_d]; marker = "s", s = 100, color = "C$(i-1)")
end
legend(); xlabel("embedding dimension")
ylabel("estimator")
tight_layout()
```

image

Datseris:
Okay, so in the `cao_embed` function I've changed the condition:
```
if y[i] ≤ thres && rat[i] > 0.5
```
from the original:
```
if y[i] ≤ thres
```

So now it not only checks the local slopes, but also requires that the Cao statistic has value greater than 0.5. It seems to work:

image

Datseris:
Tests pass now, but there is one problem remaining: the automated `ifnn` method returns embedding dimension 4 for a sinusoidal signal. I'd expect that the dimension should be 2, at most 3, but 4 is too high.

Datseris:
Anyway, I'm merging this to move on to other fixes, and get out the deprecations as soon as possible.

PR #60, opened 2020-10-22
bug_fix in pairwise distances

hkraemer:
I found a tiny bug, when I compared the L-statistic to my Matlab and Python implementations. It turns out that values from
`fiducial_pairwise_dist_sqrd` were twice as big as they should be. Tiny correction.


hkraemer:
Maybe I'll have to adapt the tests slightly. I will see in some minutes. If tests go through please merge.

PR #63, opened 2020-11-17
added pecuzal method

hkraemer:
I added the PECUZAL method incl. tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.14.0
+* New unified embedding method `pecuzal` by Kraemer et al.
+
 # v1.13.0
 * `reconstruct` is deprecated in favor of `embed`.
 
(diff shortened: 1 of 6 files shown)


Datseris:
I'll review once tests pass :)

hkraemer:
Hey @Datseris ,
fair enough.. these tests.. grrr! There you go ;)

Datseris:
@hkraemer This is in and tagged as v1.14!

If you want to write some documentation and example for it in DynamicalSystems.jl this is of course welcomed. In principle what you have to do is edit the file https://github.com/JuliaDynamics/DynamicalSystems.jl/blob/master/docs/src/embedding/estimate.md  . This is a markdown file that incorporates runnable Julia code. Everything inside a ```@example code block is runnable Julia code that will be evaluated and shown in the docs.

If there isn't a specific example you want to add it is fine, we simply add `pecuzal_embedding` in the ```@docs code block so that its documentation string is expanded.

hkraemer:
Alright, I will do that. One more question: If an example code block returns a figure I'd like to show in the docs right underneeth the code-block, how can I ensure that this is displayed? In your example with the `optimal_traditional_de`-method, you wrote ```fig.savefig("estimateD.png"); nothing # hide```. So this Figure is saved and the line of code hided from getting displayed in the docs, but does that automatically mean, that the figure gets displayed? 

Another question is how the Refs. are handled. On the very bottom of the documentation page all the Refs. from the doc-strings of the mentioned function get displayed. It this automated?

Datseris:
The refs are automated yes provided you do this [^Kraemer] thing.

For the figure, as you will ntoice, after the code block ends with the `fig.savefig` command, then next markdown command displays the figure, e.g. ``

PR #66, opened 2020-11-29
Horizontal concatenation of datasets, plus `Dataset` convenience constructors

kahaaga:
I often find myself wanting to horizontally concatenate datasets with datasets, datasets with a real-valued vector, or a real-valued vector with a dataset. 

Instead of having to redefine `Base.hcat(x::Dataset, y::Dataset)`, `Base.hcat(x::Vector{<:Real}, y::Dataset)`, and `Base.hcat(x::Dataset, y::Vector{<:Real})`, as well as `Dataset(x::Dataset, y::Dataset)`, `Dataset(x::Vector{<:Real}, y::Dataset)` and `Dataset(x::Dataset, y::Vector{<:Real})`, to achieve this, it would be nice to have this as part of DelayEmbeddings. 

This PR implements precisely that, and adds some tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# v1.15.0
+* Horizontal concatenation of same-length `Vector{<:Real}` and `Dataset` in any order using 
+  `Base.hcat(x, y)` or `[x y]` syntax.
+* Convenience constructors that uses horizontal concatenation: 
+  `Dataset(::Dataset, ::Vector{<:Real})`, `Dataset(::Vector{<:Real}, ::Dataset)` and 
+  `Dataset(::Dataset, ::Dataset)`.
+
 # v1.14.0
 * New unified embedding method `pecuzal` by Kraemer et al.
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DelayEmbeddings"
 uuid = "5732040d-69e3-5649-938a-b6b4f237613f"
 repo = "https://github.com/JuliaDynamics/DelayEmbeddings.jl.git"
-version = "1.14.4"
+version = "1.15.0"
 
 [deps]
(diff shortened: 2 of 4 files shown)


Datseris:
sure! You haven;t defined `Base.hcat(::Dataset, ::Vector)` in this PR. Also please increment minor version and add changelog entry :)

kahaaga:
>  You haven;t defined Base.hcat(::Dataset, ::Vector) in this PR. 

You're right, the `Base.hcat(::Dataset, ::Vector)` method existed before. Methods in this PR completes all two-argument combinations of horizontal concatenation between `::Dataset` and `Vector`.

> Also please increment minor version and add changelog entry :)

Done.

Datseris:
I'm really not sure about this:

```
Convenience constructors that uses horizontal concatenation: 
  `Dataset(::Dataset, ::Vector{<:Real})`, `Dataset(::Vector{<:Real}, ::Dataset)` and 
  `Dataset(::Dataset, ::Dataset)`.
```
What's the purpose? Given that `hcat` is smaller to type, why would we have two different syntaxes to do the same thing?

kahaaga:
> What's the purpose? Given that hcat is smaller to type, why would we have two different syntaxes to do the same thing?

The purpose is simply that it is more intuitive to me to do `Dataset(::Dataset, ::Dataset)`/`Dataset(::Vector{<:Real}, ::Dataset)`, `Dataset(::Dataset, ::Vector{<:Real})` when I want a new, expanded dataset that is a combination two datasets/vectors. 

`Dataset(::Vector{<:Real}, ::Vector{<:Real}, ...)` already works to combine single-variable data, so I always get surprised that I can't combine multivariate data.

If you don't think it belongs here, I can just remove the constructors. `hcat` will also work.

Datseris:
I guess you have a point about `Dataset(Vector, Vector, ...)`.

PR #67, opened 2020-11-30
Allow weights in generalized embeddings

Datseris:
Allows each entry of a `genembed` to be further multiplied by a provided weight `w`. (optional, by default is off)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,18 @@ jobs:
       matrix:
         version:
           - '1'
-        os:  [ubuntu-latest]
+        os:  [ubuntu-latest] # adjust according to need, e.g. os: [ubuntu-latest] if testing only on linux
         arch:
           - x64
     steps:
+      # Cancel ongoing CI test runs if pushing to branch again before the previous tests
+      # have finished
+      - name: Cancel ongoing test runs for previous commits
+        uses: styfle/cancel-workflow-action@0.6.0
+        with:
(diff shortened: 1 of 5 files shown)


Datseris:
@hkraemer  and @kahaaga  I'm tagging you here because this could lead to new research possibilities for you two. I know in the literature weighted traditional delays have been done, but I'm not sure whether they have been done in advanced settings like e.g. transfer entropies or unified optimal embeddings. You might want to play around with this new method in your own work.

kahaaga:
Let 

```julia
x = [1, 2, 3, 4, 5, 6, 7, 8]
js = (1, 1, 1)
τs = (0, 1, 2)
```

Then I get

```julia
julia> genembed(x, τs, js; ws = (1, 1, 1))
3-dimensional Dataset{Int64} with 6 points
 1  2  3
 2  3  4
 3  4  5
 4  5  6
 5  6  7
 6  7  8
```

Using integer weights works: 

```julia
julia> genembed(x, τs, js; ws = (1, -20, 5))
3-dimensional Dataset{Int64} with 6 points
 1   -40  15
 2   -60  20
 3   -80  25
 4  -100  30
 5  -120  35
 6  -140  40
```

But weights using floats do not:

```julia
julia> genembed(x, τs, js; ws = (0.5, 0.5, 1.0))
ERROR: InexactError: Int64(0.5)
```

Is this intended behaviour?

Datseris:
nope, it is obviously a bug :D I'll fix it and add it as test case!

PR #69, opened 2020-11-30
Weights docs

kahaaga:
Sorry, I should have paid more attention! I missed a typo in PR #67. 

Also, I added a small clarification to the docstring about the weights, because it was a bit unclear to me what the `ws` argument is supposed to do reading the docstrings the first few times (I had to experiment with `genembed` to understand what the `ws` argument did). I think it is a bit more clear now. However, if you feel it's better the way it is, I can remove the additional explanation it.


Datseris:
perfect thanks

PR #73, opened 2020-12-01
redefine Dataset indexing

### LINKED ISSUE #72: Return type of one-dimensional Datasets slices
In other places (e.g in https://github.com/JuliaDynamics/RecurrenceAnalysis.jl/issues/106) we have been discussing why multi-dimensional time series should be using the `Dataset` type instead of `AbstractMatrix`.

I think that one of the selling points of `Dataset` is that it can be used to iterate over the multi-dimensional points just as it is done with vectors, e.g. if you have a series of 3-dimensional points, you can have them in a dataset `data` and iterate over those points just with `for x in data`. On the other hand, with `AbstractMatrices` you have to take into account if the points are organized in rows or columns, and then do `for x in eachrow(data)` etc.

However that argument would be stronger if one-dimensional slices like `data[1:10]` also returned a `Dataset`. As it is now, that returns a vector of `SArray`, and the straightforward way of slicing `Dataset`s is with matrix-like indexing, i.e. `data[1:10, :]`. This is not good for generic code that would like to manage `Dataset`s more or less like `Vector`s. (`v[1:10, :]` for vectors returns a matrix!)

Would it be sensible to rewrite the method `Base.getindex(d::AbstractDataset, r::AbstractRange)` to achieve that, or would it be too breaking?

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/99991596-return-type-of-one-dimensional-datasets-slices?utm_campaign=plugin&utm_content=tracker%2F115474986&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F115474986&utm_medium=issues&utm_source=github).
</bountysource-plugin>

heliosdrm:
Fixes #72

### DIFF
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -142,7 +142,8 @@ In the following let `i, j` be integers,  `typeof(data) <: AbstractDataset`
 and `v1, v2` be `<: AbstractVector{Int}` (`v1, v2` could also be ranges).
 
 * `data[i]` gives the `i`th datapoint (returns an `SVector`)
-* `data[v1]` will return a vector of datapoints
+* `data[v1]`, where `v1` is a collection of indices, returns a `Dataset`
+  with the points in those indices.
 * `data[v1, :]` using a `Colon` as a second index will return a `Dataset` of
   these points
 * `data[:, j]` gives the `j`th variable timeseries, as `Vector`
--- a/src/dataset.jl
+++ b/src/dataset.jl
@@ -29,7 +29,8 @@ Base.eachcol(ds::AbstractDataset) = (ds[:, i] for i in 1:size(ds, 2))
 Base.eachrow(ds::AbstractDataset) = ds.data
 
 # 1D indexing over the container elements:
-@inline Base.getindex(d::AbstractDataset, i) = d.data[i]
+@inline Base.getindex(d::AbstractDataset, i::Int) = d.data[i]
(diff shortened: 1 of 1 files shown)


Datseris:
Thanks, the docstring entry also has to be changed to reflect this.

I will merge this PR last just before I'm ready to get 2.0 out, because this change (unfortunately) cannot be deprecated.

heliosdrm:
Hopefully, although it does break the declared API, it won't be too disruptive. At least all the tests of the package passed without changes. I wonder how many other packages actually rely on `data[v1]` returning a vector of `SArray`s and not a `Dataset`.

src/dataset.jl
```diff
@@ -142,7 +142,8 @@ In the following let `i, j` be integers,  `typeof(data) <: AbstractDataset`
 and `v1, v2` be `<: AbstractVector{Int}` (`v1, v2` could also be ranges).
 
 * `data[i]` gives the `i`th datapoint (returns an `SVector`)
-* `data[v1]` will return a vector of datapoints
+* `data[v1]`, where `v1` is a collection of indices, returns a `Dataset`
+  with the points in those indices.
 * `data[v1, :]` using a `Colon` as a second index will return a `Dataset` of
   these points
```
Datseris, line 148:
```suggestion
* `data[v1] == data[v1, :]`, returns a `Dataset` with the points in those indices.
```

PR #75, opened 2020-12-06
deprecate mutualinformation

Datseris:
This PR deprecates the name `mutualinformation` because it will result to both conflict and confusion. For a long time I've been wanting to add a proper mutual information method `mutualinfo(x, y; ...)` in one of the packages of JuliaDynamics.

This function missleads with its name, because it only works for _self_ - mutual information, and I don't see a straightforward way to extend it to new arguments.

The new function `mutualinfo` will serve this purpose. At the moment we are discussing whether this function should go into Entropies.jl or TransferEntropy.jl.

### DIFF
--- a/src/traditional_de/estimate_delay.jl
+++ b/src/traditional_de/estimate_delay.jl
@@ -14,7 +14,7 @@ The `method` can be one of the following:
 * `"ac_min"` : delay of first minimum of the auto-correlation function.
 * `"mi_min"` : delay of first minimum of mutual information of `s` with itself
   (shifted for various `τs`).
-  Keywords `nbins, binwidth` are propagated into [`mutualinformation`](@ref).
+  Keywords `nbins, binwidth` are propagated into [`selfmutualinfo`](@ref).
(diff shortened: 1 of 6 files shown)


src/traditional_de/estimate_delay.jl
```diff
@@ -119,11 +119,16 @@ end
 #####################################################################################
 #                               Mutual information                                  #
 #####################################################################################
-export mutualinformation
+export selfmutualinfo, mutualinformation
+function mutualinformation(args...; kwargs...)
+    @warn "`mutualinformation` is deprecated in favor of `selfmutualinfo`."
```
heliosdrm, line 124: This might go into a different file: `deprecate.jl`. I like having deprecated functions methods in a different place, because it helps other people to have a clear perspective of the situation.

heliosdrm (approved):

src/traditional_de/estimate_delay.jl
Datseris: sure!

PR #76, opened 2020-12-07
implement SubDataset

### LINKED ISSUE #74: Could we define "views" of `Dataset`s?
The only way of taking a fragment of a `Dataset` is creating another one. This is not optimal, and it might be good to have another subtype of `AbstractDataset`, whose contents are a view to the data of another `Dataset`.

The easiest would be to define:

```julia
Base.view(d::AbstractDataset, i) = view(d.data, i)
```
and then some type like `DatasetView <: AbstractDataset` whose data is a `SubArray{<:SArray}`, such that one could do:

```julia
d = Dataset(x)
dv = Dataset(@view d[xx])
```
(Notice that this code actually works right now, but it produces a copy of the contents before creating `dv`.)

Such `dv::DatasetView` should behave as if I had written `dv = d[xx]`, but that way we would save the superfluous allocation.

I acknowledge that it would be nicer if the interface is directly `dv = @view d`, but I'm not sure if that would be right. `Base.view` (and thus the macro `@view`) is meant to produce a `SubArray`, which is defined as a [container encoding a "view" of a parent `AbstractArray`](https://docs.julialang.org/en/v1/devdocs/subarrays/). So, although technically it might be feasible to write a method of `Base.view` that takes and returns an `AbstractDataset`, for me that would only feel right if we define `AbstractDataset <: AbstractArray` -- which maybe makes sense too, but might need some deeper thought to ensure that the API of `AbstractDataset` is fully consistent with what is expected from `AbstractArray`s (or at least `AbstractVector`s).

heliosdrm:
Closes #74 

The new `SubDataset` type contains a reference not only to the subarray of data, but to the parent dataset, such that the function `parent` can also be extended to this type. 

Two doubts:

1. Only one index is allowed for views, but `view(dataset, i, :)` (taking all the "columns") could be possible, and equal to `view(dataset, i)`. Shall we allow it, or leave it?
2. I've made an inner `SubDataset` constructor that checks that the parent of the subarray is the data vector contained in the parent. But it might be removed, if we consider that such check is not needed.

### DIFF
--- a/src/subdataset.jl
+++ b/src/subdataset.jl
@@ -1,0 +1,41 @@
+"""
+    SubDataset{D, T, <:AbstractDataset{D,T}, <:SubArray{<:SVector{D,T},1}} <: AbstractDataset{D,T}
+
+A view of an `AbstractDataset`, as returned by the `view` function
+or the `@view` macro. A `SubDataset` is an `AbstractDataset`
+of the same type as its parent, so indexing and most functions
+can be expected to work in the same way for both the parent and the view.
+"""
+struct SubDataset{D, T, P<:AbstractDataset{D,T}, S<:SubArray{<:SVector{D,T},1}} <: AbstractDataset{D,T}
+    parent::P
+    data::S
+    function SubDataset(par, data)
+        @assert parent(data) === par.data
+        P = typeof(par)
+        S = typeof(data)
+        SV = eltype(P)
+        T = eltype(SV)
+        D = length(SV)
+        new{D,T,P,S}(par, data)
+    end
+end
+
+function Base.summary(sd::SubDataset{D, T}) where {D, T}
+    N = length(sd)
+    return "$N-element view of $D-dimensional Dataset{$(T)}"
+end
+
+Base.parent(sd::SubDataset) = sd.parent
+Base.parentindices(sd::SubDataset) = parentindices(sd.data)
+
+"""
+    view(d::Dataset, indices)
+
+Return a view into the parent dataset `d`, as a [`SubDataset`](@ref)
+that contains the datapoints of `d` referred to by `indices`.
+"""
+Base.view(d::AbstractDataset, i) = SubDataset(d, view(d.data, i))
+
+function Base.view(::AbstractDataset, ::Any, ::Any, ::Vararg)
+    throw(ArgumentError("Dataset views only accept indices on one dimension"))
+end
--- a/test/dataset_tests.jl
+++ b/test/dataset_tests.jl
@@ -31,6 +31,7 @@ println("\nTesting Dataset...")
     sub = @view data[11:20]
     @test sub isa DelayEmbeddings.SubDataset
     @test sub[2] == data[12]
+    @test dimension(sub) == dimension(data)
   end
 
   @testset "minmax" begin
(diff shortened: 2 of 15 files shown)


src/subdataset.jl
```diff
@@ -0,0 +1,41 @@
+"""
+    SubDataset{D, T, <:AbstractDataset{D,T}, <:SubArray{<:SVector{D,T},1}} <: AbstractDataset{D,T}
+
+A view of an `AbstractDataset`, as returned by the `view` function
+or the `@view` macro. A `SubDataset` is an `AbstractDataset`
+of the same type as its parent, so indexing and most functions
+can be expected to work in the same way for both the parent and the view.
+"""
+struct SubDataset{D, T, P<:AbstractDataset{D,T}, S<:SubArray{<:SVector{D,T},1}} <: AbstractDataset{D,T}
+    parent::P
```
Datseris, line 10: What is the benefit of having a parent? For the code here I don't see it being used anywhere. If it is useful then okay, but if it is not then we should remove it since it increases compile time.

test/dataset_tests.jl
```diff
@@ -27,6 +27,10 @@ println("\nTesting Dataset...")
     @test size(data[1:10,1:2]) == (10,2)
     @test data[1:10,1:2] == Dataset(a[1:10], b[1:10])
     @test data[SVector{10}(1:10), SVector(1, 2)] == data[1:10, 1:2]
+
+    sub = @view data[11:20]
+    @test sub isa DelayEmbeddings.SubDataset
+    @test sub[2] == data[12]
```
Datseris, line 33: could you please add one more test that checks `dimension(sub) == dimension(data)` ?

src/subdataset.jl
heliosdrm: The benefit is consistency with the API of views for other types, which allow to retrieve the parent array (or data frame, etc.) with the function `parent`.

I have not used it myself so far, but I can imagine some use case where it might be useful. For instance, a function that takes a view of a dataset, does something with it, and returns another view into the same dataset "moving the window" by some stride, to help some iterative workflow. To do that you do need the parent dataset (plus the indices of the view, which are available with `parentindices`).

(That particular example might be worked around by using the parent of the `SubArray` contained in the `data` field. But it feels hackish, and the parent obtained that way would be a vector of `SVector`, not an `AbstractDataset`.)

test/dataset_tests.jl
heliosdrm: done in 8978624

src/subdataset.jl
Datseris: I'm convinced !

Datseris:
Seems good to me! Only thing that remains is incrementing `Project.toml` and adding an entry in the CHANGELOG

heliosdrm:
> Seems good to me! Only thing that remains is incrementing `Project.toml` and adding an entry in the CHANGELOG

Not sure about how to update the CHANGELOG, which as in master is inconsistent with the actual history: deprecation of `mutualinformation` is marked as a change of v1.18.0, which does not exist yet!

Datseris:
Well then you can put this under `1.18` as well, so that we release both of these changes together...? I simply forgot to tag a release.

heliosdrm:
Moved the line from CHANGELOG that announced the deprecation of `mutualinformation` to v1.17, where it belongs according to the real history.

Datseris:
aaaaaaaaaa now I get it, okay sorry :D

Datseris:
I'll go through the rest of the library, e.g. estimating optimal embeddings, to see if we use dispatch on `::Dataset` (incorrect) instead of `::AbstractDataset` (correct).

Datseris:
Okay I think I've updated everything. I'll let the test go, hopefully I didn't break anything, and then I'll merge.

PR #78, opened 2020-12-20
Improve is stochastic clause for automated TDE

Datseris:
@hkraemer

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+# v1.19.3
+* Automated traditional delay embedding has improved clause for Cao's method.
 # v1.19.0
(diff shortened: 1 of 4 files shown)


Datseris:
At the moment we just print a statement "Embedding not achieved." Shouldn't we error instead?

hkraemer:
Jepp, we should return the time series, which has been put in, in this case and leave the message, I suggest.

Datseris:
I'm asking whether the function should **error** instead of simply **printing** a message.

hkraemer:
I suggest to let it print that message and return the single time series. Might be more handy in a loop-evaluation

PR #79, opened 2021-01-18
Bugfix hcat lagged

hkraemer:
Here is a minor bugfix for the function `hcat_lagged_values`. The former version always reduced the length of the output Dataset by the given delay value `τ`. Now it is just reducing the size according to the available data in `s`. @Datseris , you made it right in the implementation of `genembed` though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+# v1.20.0
+* Revised version of PECUZAL algorithm, now tracking maximal L-decrease
+
 # v1.19.3
 * Automated traditional delay embedding has improved clause for Cao's method.
 # v1.19.0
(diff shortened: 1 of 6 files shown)


Datseris:
Alright, thanks. We need to add a basic test for this function, given that we already made a mistake with it once. I don't think `hcat_lagged_values` is used anywhere in the tests.

hkraemer:
No it is not. I will incorporate it when adjusting the tests to the new PECUZAL version. This bugfix, however, might break the tests of the actual PECUZAL tests...

Datseris:
You can do both in this PR.

Datseris:
Hi @hkraemer , please increment minor version in Project.toml, and also add an entry in CHANGELOG.md with a one-sentence description of changes users should be aware of. Then I'll merge and tag a new version.

PR #80, opened 2021-01-19
Bug fix pecuzal kwa

hkraemer:
Hey @Datseris ,
there was a bug in one keyword-argument for the multivariate case (`Datset`-input). The reason why I did not see is was that I also used the "wrong" keyword in the tests... It is consistent now with the docstring of the function.


Datseris:
Please increment patch version in Project.toml so we can release immediatelly.

PR #81, opened 2021-03-13
Bug fix pecuzal

hkraemer:
Hey @Datseris ,
here is a minor bug-fix. It is simply a raised exception, when there are `NaN`-values, due to 0-distances from the nearest neighbour-search. I encountered this, while playing around with perfect, noise-free model-data.
Cheers

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "DelayEmbeddings"
 uuid = "5732040d-69e3-5649-938a-b6b4f237613f"
 repo = "https://github.com/JuliaDynamics/DelayEmbeddings.jl.git"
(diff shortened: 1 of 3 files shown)


Datseris:
Can you increment patch version in `Project.toml` as well?

---

In what kind of "model data" does this happen? I thought we have tested this algorithm with "model data" from simple models like Roessler.

hkraemer:
It happend on a model for EEG-Data, which produced a lot of "same values". It doesn't happen for Roessler or any low dimensional chaotic system. Can also happen on artifical sinusodials for instance. As soon as you add `1e-10*rand()` or so to it, it is all good.

Issue #96, opened 2021-08-27
Normalization factor in Uzal-costfunction might be wrong.


hkraemer:
Hey all,

take a look [here](https://github.com/hkraemer/PECUZAL_Matlab/pull/2#issue-974391197). This seems to be valid. The correction also follows from Eqs. (17), (18) & (20) in the [original publication](https://journals.aps.org/pre/abstract/10.1103/PhysRevE.84.016223). Thanks to @JayeshMD , who digged deeper into the math. However, this will change the general offset of the `L`-function and Fig.7 in the original paper would look qualitatevily the same (even better), but the graphs have a different offset. The change in the code is subtle, but it does make sense from a physical perspective I'd say. I also tried around with `L`-functions of different time series lengths of the same system and got similar results as @JayeshMD has shown in his [slides](https://github.com/hkraemer/PECUZAL_Matlab/pull/2#issue-974391197): With high time series lengths the value of `L` saturates/converges. This does not happen the way it is indicated in the paper and also implemented here so far. 

Another remark regarding PECUZAL: PECUZAL is not affected by these changes, since there we look at ratios of `L` for adjacent embedding dimensions and, thus, the time series length cancels out.

I am already preparing a PR and will adjust the tests, which I compare with the outcomes of the PECUZAL-implementations in [MATLAB](https://github.com/hkraemer/PECUZAL_Matlab) and [Python](https://github.com/hkraemer/PECUZAL_python). But I can not use the Figures in the original publication as a benchmark anymore. So let me know, if you are fine with that. While doing that I will also adjust the tests for PECUZAL, regarding #95 .

Cheers

Issue #95, opened 2021-08-01
Unified embedding tests are brittle


Datseris:
We have established the validity of our unified embedding methods because we even wrote a paper utilizing them in detail and comparing them with each other. There isn't any doubt that they work. However, the way we have written our tests is brittle, because (due to chaos) running the test on different machines may make them fail. For example while working on PR #93 I've seen things like:

```
Multivariate example: Test Failed at c:\Users\m300808\.julia\dev\DelayEmbeddings\test\unified\test_pecuzal_embedding.jl:72
  Expression: -0.356 < Ls[2] < -0.355
   Evaluated: -0.356 < -0.32831887625247136 < -0.355

Testing pecuzal_method.jl...
Initializing PECUZAL algorithm for univariate input...
Starting 1-th embedding cycle...
Starting 2-th embedding cycle...
Starting 3-th embedding cycle...
Algorithm stopped due to increasing L-values. VALID embedding achieved ✓.
  5.931637 seconds (15.70 M allocations: 2.972 GiB, 14.40% gc time)
Univariate example: Test Failed at c:\Users\m300808\.julia\dev\DelayEmbeddings\test\unified\test_pecuzal_embedding.jl:19
  Expression: -0.728 < Ls[1] < -0.727
   Evaluated: -0.728 < -0.7839819804977819 < -0.727
```

and even

```
Initializing PECUZAL algorithm for multivariate input...
Starting 1-th embedding cycle...
Starting 2-th embedding cycle...
Starting 3-th embedding cycle...
Algorithm stopped due to increasing L-values. VALID embedding achieved ✓.
 21.312898 seconds (71.51 M allocations: 11.922 GiB, 8.30% gc time)
Multivariate example: Test Failed at c:\Users\m300808\.julia\dev\DelayEmbeddings\test\unified\test_pecuzal_embedding.jl:55
  Expression: length(ts_vals) == 5
   Evaluated: 3 == 5
```

(i.e. a lower-dimensional embedding is produced sometimes)

We need to re-write tests in a different manner. First increase the error margins, but also come up with tests that are not so brittle and depend sensitively on the chaotic nature of trajectories. Furthermore, this issue needs to be solved in conjuction with #94 , which would make trajectories no longer simulated but rather available as `.csv` files from the get go.

@hkraemer I'm also cc'ing you for reference.

Datseris:
Unfortunately PECUZAL is the most brittle, seems to have the most failures when running on different PC.

[...]

Issue #94, opened 2021-07-31
Resolve circular dependencies


Datseris:
DynamicalSystemsBase.jl depends on DelayEmbeddings.jl. However DelayEmbeddings.jl uses DynamicalSystemsBase.jl in its test suite. That creates problems when updating major versions (in fact it makes them impossible on paper).

The only way to resolve this is to not use DSB.jl in DE.jl's test. One solution is to write all trajectories needed into CSV files and save them in the main JuliaDynamics repo that holds the docs, and download them each time tests are run (we already do this for some trajectories).

PR #97, opened 2021-08-31
New tests for pecuzal, uzal_cost; corrected normalization in uzal cost

hkraemer:
This is regarding #95, #96 and also #94. The correction in the computation of Uzal's cost function have been implemented and the according tests have been revised. Same holds for the brittle PECUZAL-tests. In the PECUZAL case I double-checked each test with the outcomes from my [Matlab](https://github.com/hkraemer/PECUZAL_Matlab) and [Python](https://github.com/hkraemer/PECUZAL_Python) implementations of the algorithm.  I also relaxed the test-constraints and outsourced the time series being used (we should continuing doing this for the other methods). Hopefully this works now. Cheers


test/runtests.jl
```diff
@@ -12,6 +14,13 @@ for a in todownload
     download(repo*"/"*a, joinpath(tsfolder, a))
 end
 
+#New:
+repo = "https://raw.githubusercontent.com/JuliaDynamics/JuliaDynamics/master/timeseries"
+a = "test_time_series_lorenz_standard_N_10000_multivariate.csv"
+download(repo*"/"*a, joinpath(tsfolder, a))
+a = "test_time_series_roessler_N_10000_multivariate.csv"
+download(repo*"/"*a, joinpath(tsfolder, a))
+
```
Datseris, line 23: Should be written in terms of a loop. Define `files = [all file names here]` and loop over `for a in files`

hkraemer:
hmpf, I think I PR'ed from an old branch... hm. Might not be a problem hopefully.

Datseris:
it's okay, no conflicts. tests pass, do you need to add anything more here or the PR should be merged?

hkraemer:
no I think I'll do the loop thingy and that's it

hkraemer:
I'll fix the other two test next week or so. Super busy atm :(

Datseris:
No problem, take your time. Don't forget that we need to tag a new release for these changes to be propagated beyond development in a stable release. You should tell me when you feel a new version should come out and I'll tag it.

PR #99, opened 2021-12-01
try to fix 1.7

Datseris:
cc @kahaaga


kahaaga:
@Datseris Perfect! Thanks for the fix.

PR #101, opened 2022-01-18
Time series for testing

hkraemer:
Similar to the procedure in `RecurrenceAnalysis.jl` and the discussion in #94 the timeseries for testing are now stored in one repo.


hkraemer:
This is not a bug-fix, so I did not increment the version.patch

PR #103, opened 2022-02-23
Bug fix pecora samplesize

hkraemer:
I encountered a whole lot of useless code-snippets in `pecora.jl`. In particular the `samplesize`-keyword did not do anything. It was is simple fix and now it works (considerably fast actually). I also adjusted the tests with respect to that change and got rid of the `Peaks`-dependency.


Datseris:
If we don't need the `Peaks` dependency, then you need to also remove it from Project.toml, and remove any leftover `using Peaks` isntances.

hkraemer:
It is still used in `test_garcia`, do you want me to replace that as well in this PR?

Datseris:
yeah that would be lovely

Issue #106, opened 2022-05-02
Make `Matrix` available as return type of `orthonormal`


ikottlarz:
# Issue description
The `orthonormal` function that is also used in `lyapunovspectrum` over in `ChaosTools.jl` always returns a `SMatrix`. This can lead to issues with high-dimensional systems, as 
> A very rough rule of thumb is that you should consider using a normal Array for arrays larger than 100 elements.
( [Readme of `StaticArrays.jl`](https://github.com/JuliaArrays/StaticArrays.jl#speed))

The return type should be adjustable depending on the system size (see [this issue](https://github.com/JuliaDynamics/ChaosTools.jl/issues/255) in `ChaosTools.jl`)

PR #107, opened 2022-05-02
added option for different return type for orthonormal function

ikottlarz:
I think this code is not very pretty, I haven't seriously coded in `julia` in a while, so feel free to tell me what I could/ should do better. But I think it works at least. 
This would solve #106

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -109,7 +109,7 @@ Return a matrix `ws` with `k` columns, each being
 an `D`-dimensional orthonormal vector.
 `T` is the return type and can be either `SMatrix` or `Matrix`
 
-returns either SMatrix{D, k} if D*k < 100, otherwise Matrix{D, k}
+returns either SMatrix{D, k} if D*k < 100, otherwise Dxk Matrix{Float64}
 """
 function orthonormal end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -108,7 +108,7 @@ end
 Return a matrix `ws` with `k` columns, each being
 an `D`-dimensional orthonormal vector.
 
-Default: returns SMatrix{D, k} if D*k < 100, otherwise Matrix
+returns either SMatrix{D, k} if D*k < 100, otherwise Matrix{D, k}
 """
 function orthonormal end
 
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -104,19 +104,25 @@ function to_vectorSvector(a::AbstractMatrix)
 end
 
 """
-    orthonormal(D, k) -> ws
+    orthonormal([T,] D, k) -> ws
 Return a matrix `ws` with `k` columns, each being
 an `D`-dimensional orthonormal vector.
 
-Always return `SMatrix` for stability reasons.
+Default: returns SMatrix{D, k} if D*k < 100, otherwise Matrix
 """
 function orthonormal end
 
-@inline function orthonormal(D::Int, k::Int)
-    k > D && throw(ArgumentError("k must be ≤ D"))
-    q = qr(rand(SMatrix{D, k})).Q
-end
+orthonormal(D, k) = D*k < 100 ? orthonormal(SMatrix, D, k) : orthonormal(Matrix, D, k)
 
+@inline function orthonormal(T::Type, D::Int, k::Int)
+    k > D && throw(ArgumentError("k must be ≤ D"))
+    if T == SMatrix
+        q = qr(rand(SMatrix{D, k})).Q
+    elseif T == Matrix
+        q = Matrix(qr(rand(Float64, D, k)).Q)
+    end
+    q
+end
 
 """
     hcat_lagged_values(Y, s::Vector, τ::Int) -> Z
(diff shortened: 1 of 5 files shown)


src/utils.jl
```diff
@@ -104,20 +104,26 @@ function to_vectorSvector(a::AbstractMatrix)
 end
 
 """
-    orthonormal(D, k) -> ws
+    orthonormal([T,] D, k) -> ws
 Return a matrix `ws` with `k` columns, each being
 an `D`-dimensional orthonormal vector.
 
-Always return `SMatrix` for stability reasons.
+Default: returns SMatrix{D, k} if D*k < 100, otherwise Matrix
```
Datseris, line 111: From this documentation string it is not clear what `T` should be.

src/utils.jl
```diff
@@ -104,20 +104,26 @@ function to_vectorSvector(a::AbstractMatrix)
...
 """
 function orthonormal end
 
-@inline function orthonormal(D::Int, k::Int)
+orthonormal(D, k) = D*k < 100 ? orthonormal(SMatrix, D, k) : orthonormal(Matrix, D, k)
```
Datseris, line 115: This is a breaking change. But to be honest, this is definitely worth doing... But please mention this in the documentation string. What you say now is not true, thje default is not to return `SMatrix`, because the default depends on the input.

Datseris (commented):
> , I haven't seriously coded in julia in a while

Where have you been coding then??? I hope you are not living a really sad life in python land.

This looks okay, but the docstring needs some improvements still. Furthermore you also need to add tests (make a test that makes a 50x50 orthonormal), increase minor version in Project.toml, and add an entry in the changelog.

src/utils.jl
ikottlarz: So, breaking change meaning increasing the major version? I'm confused..

src/utils.jl
```diff
@@ -108,7 +108,7 @@ end
 Return a matrix `ws` with `k` columns, each being
 an `D`-dimensional orthonormal vector.
 
-Default: returns SMatrix{D, k} if D*k < 100, otherwise Matrix
+returns either SMatrix{D, k} if D*k < 100, otherwise Matrix{D, k}
```
Datseris, line 112: Notice that in Julia matrices don't have known size by type. There is no `Matrix{D, k}`. The only thing that exists is `Matrix{T}` with `T` the element type.

test/utils_test.jl
```diff
@@ -0,0 +1,10 @@
+using Test, DelayEmbeddings
+
+println("\nTesting utils...")
+
+@testset "utils" begin
+
+    A = orthonormal(50, 50)
+    @test size(A) == (50, 50)
```
Datseris, line 8: Wait, the `orthonormal` function was not tested anywhere else so far...? In anycase, please also add 2 more tests, one that checks that `A isa Matrix`, and another one `B = orthonormal(3,3)` and `@test B isa SMatrix`.

Datseris:
> You know, with properly starting to count at 0 and all...

of course, because it makes perfect sense that the 2nd element of an ordered collection having index 1, and the 64th element having index 63. That's exactly how we do it in mathematics, science, and every day life as well if I am not mistaken.

test/utils_test.jl
ikottlarz: Actually, the function was neither tested nor used (anymore?) in this repository, I did a keyword search through all files and it only came up in its definition in `utils.jl`. But yes, good point, I'll write those tests in a bit once I'm at my PC

ikottlarz:
> of course, because it makes perfect sense that the 2nd element of an ordered collection having index 1, and the 64th element having index 63. That's exactly how we do it in mathematics, science, and every day life as well if I am not mistaken.

Well, my first programming language(s) was/ were C/ C++, so with all those pointers in mind, starting to count at 0 makes a lot more sense to me :-)

src/utils.jl
ikottlarz: Ah, yes, I even realized that when I wrote the code. Changed it, sorry!

src/utils.jl
ikottlarz: Made a comment about that, should be clearer now

Datseris:
@ikottlarz can you do me a favor and check if your `lyapunovspectrum` problem is solved with this PR?

ikottlarz:
> can you do me a favor and check if your lyapunovspectrum problem is solved with this PR?

@Datseris I'm in the process of doing this right now. So far, it looks good, and for now I can say that it at least doesn't break at the same system size as before (broke at `dimension(ds) = 270` with `k = 32` , i.e. a `270x32` Matrix, I'm currently at `dimension(ds) = 320` with `k = 37` and no error yet). I think it solves my error. But I'm calculating the spectrum up to a `dimension(ds) = 500` with `k = 55` over night anyway, so I can also confirm more surely tomorrow  morning if you want

ikottlarz:
> so I can also confirm more surely tomorrow morning if you want

... okay, so this gets really slow, but I'm currently at `dimension(ds) = 450` with `k=50` and so far everything has worked. I'm pretty sure that particular bug is fixed now

Datseris:
thank you!

PR #108, opened 2022-08-14
Dataset distance function

Datseris:
Add a function that computes distances between datasets. Also include the Hausdorff distance.  TODO: Tests.


src/datasets/dataset_distance.jl
```diff
@@ -0,0 +1,47 @@
...
+"""
+    Hausdorff(metric = Euclidean())
+A type used in specifying dataset distances in [`dataset_distance`](@ref).
+"""
+struct Hausdorff{M<:Metric}
+    metric::M
+end
+
+function dataset_distance(d1::AbstractDataset, d2, h::Hausdorff)
+    tree1, tree2 = KDTree.((d1, d2), h.metric)
```
awage, line 42: This line gives me an error. I fixed it with 
```
    tree2 = KDTree(d2, h.metric)
    tree1 = KDTree(d1, h.metric)
```

awage:
I have tried a simple test: 
```julia
  @testset "Dataset Distances" begin
    d1 = range(1,10,length = 100) 
    d2 = d1 .+ 10
    @test dataset_distance(Dataset(d1), Dataset(d2)) == 1. 
    @test dataset_distance(Dataset(d1), Dataset(d2), Hausdorff(Euclidean())) == 10. 
  end
```
It throws an error since `dataset_distance` returns a vector for the Hausdorff version.

src/datasets/dataset_distance.jl
```diff
@@ -0,0 +1,47 @@
...
+    # This yields the minimum distance between each point
+    _, vec_of_distances12 = bulksearch(tree1, vec(d2), NeighborNumber(1))
+    _, vec_of_distances21 = bulksearch(tree2, vec(d1), NeighborNumber(1))
+    return max(maximum(vec_of_distances12), maximum(vec_of_distances21))
```
awage, line 46: This forces to return a value, but it is ugly:  
` max(maximum(vec_of_distances12), maximum(vec_of_distances21))[1]`

Issue #112, opened 2022-10-16
Merging multiple datasets in one operation


kahaaga:
Is it possible to merge multiple `Dataset`s into one larger `Dataset` using a single operation, say splatting a tuple of `Datasets` like `Dataset(ds...)`?

For example, if I have three `Dataset{2}` with the same length, I can do the following to obtain a `Dataset{6}`.

```julia
julia> x, y, z = Dataset([1:10 2:11]), Dataset([3:12 4:13]), Dataset([5:14 6:15])
julia> Dataset(Dataset(x, y), z)
6-dimensional Dataset{Int64} with 10 points
  1   2   3   4   5   6
  2   3   4   5   6   7
  3   4   5   6   7   8
  4   5   6   7   8   9
  5   6   7   8   9  10
  6   7   8   9  10  11
  7   8   9  10  11  12
  8   9  10  11  12  13
  9  10  11  12  13  14
 10  11  12  13  14  15
```

But I'd like to be able to just `Dataset(x, y, z, ...)`, potentially with more than three variables. It is of course possible to do nested calls to the `Dataset` constructor, but that leads to ugly code. The following doesn't work, but it would be nice if it did:

```julia
julia> Dataset(x, y, z)
ERROR: MethodError: no method matching Dataset(::Dataset{2, Int64}, ::Dataset{2, Int64}, ::Dataset{2, Int64})
Closest candidates are:
  Dataset(::AbstractDataset{D1, T}, ::AbstractDataset{D2, T}) where {D1, D2, T} at ~/.julia/packages/DelayEmbeddings/Lcs5l/src/datasets/dataset.jl:206
  Dataset(::Vector{<:Real}, ::AbstractDataset{D, T}) where {D, T} at ~/.julia/packages/DelayEmbeddings/Lcs5l/src/datasets/dataset.jl:208
  Dataset(::Dataset) at ~/.julia/packages/DelayEmbeddings/Lcs5l/src/datasets/dataset.jl:163
  ...
Stacktrace:
 [1] top-level scope
   @ REPL[52]:1
```

And the following creates a dataset of datasets:

```julia
julia> Dataset([x, y, z])
1-dimensional Dataset{Dataset{2, Int64}} with 3 points
 2-dimensional Dataset{Int64} with 10 points
 2-dimensional Dataset{Int64} with 10 points
 2-dimensional Dataset{Int64} with 10 points
```

Datseris:
[...]

PR #113, opened 2022-10-17
Hcat multiple datasets

kahaaga:
Addresses #112, and attempts to improve tests too (see #109)

Changes:
- `hcat`, and hence the `Dataset` constructor,  now accepts more than two datasets.
- Replaced concatenation tests with analytical tests.

### DIFF
--- a/src/datasets/dataset.jl
+++ b/src/datasets/dataset.jl
@@ -111,7 +111,6 @@ function Base.hcat(ds::Vararg{AbstractDataset{D, T} where {D}, N}) where {T, N}
     maxlen = maximum(Ls)
     all(Ls .== maxlen) || error("Datasets must be of same length")
     newdim = sum(dimension.(ds))
-    v = Vector{SVector{D, T}}(undef, maxlen)
     v = Vector{SVector{newdim, T}}(undef, maxlen)
     for i = 1:maxlen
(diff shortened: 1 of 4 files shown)


src/datasets/dataset.jl
```diff
@@ -106,13 +106,14 @@ function Base.hcat(x::Vector{<:Real}, d::AbstractDataset{D, T}) where {D, T}
...
-    L = length(x)
-    D = D1 + D2
-    v = Vector{SVector{D, T}}(undef, L)
-    for i = 1:L
-        v[i] = SVector{D, T}((x[i]..., y[i]...,))
+function Base.hcat(ds::Vararg{AbstractDataset{D, T} where {D}, N}) where {T, N}
+    Ls = length.(ds)
+    maxlen = maximum(Ls)
+    all(Ls .== maxlen) || error("Datasets must be of same length")
+    D = sum(dimension.(ds))
```
Datseris, line 113: This unecessarily restricts all datasets to have same dimension `D`. Furthermore, it replaces the type parmaeter `D` in the call signature, with a new and different number `D` later on, which is bad practice (because it is typed, otherwise it would be okay).

Given that you never actually utilize the first `D`, you can just make the call signature `AbstractDataset` without `D, T` and get `T = eltype(first(ds))`.

Datseris:
Cool, bump project version and add changelog entry and we can merge!

Issue #64, opened 2020-11-24
Error in estimate_delay when using exp_decay


pucicu:
The function `lag = estimate_delay(tr, "exp_decay", 0:100)` can throw an error of the type

`ERROR: The correlation function has elements that are ≤ 0. We can't fit an exponential to it. Please choose another method.`

It is clear what it means (and the error can be avoided when using a shorter delay interval), but in my opinion it would be better to catch the situation where auto-correlation values could be negative and to truncate the auto-correlation function to the interval of positive values.

Datseris:
Ah okay, that's a fair and straightforward improvement.

PR #122, opened 2023-03-29
Estimate delay error

denizhanpak:
Based on issue #64


denizhanpak:
T
his should be on a separate branch I believe

